Issue #18, opened 2015-12-11
Make MemDiskGroup recursive.


kiyo-masui:
Referencing a group from MemDiskGroup should return a MemDiskGroup, not a MemGroup or h5py.Group.

One thing that is subtle about this is dataset and group name restriction. This is is easy as long as both creation and referencing is handled by the root group.

kiyo-masui:
Actually, everything would have to be handled by the root group, even the `parent` attribute and `keys` method. Because we can't hold a any permanent references (only weak references), we can't count on things like `parent` to exist.  Thus root needs to generate parent when we reference the attribute.

Alternately, since there are no upward references (__get_item__ always generates a new object, in contrast to MemGroup), we probably don't have to worry about circular references in the same way.

Actually, h5py does this. If `g` is an `h5py.Group`, then `g.parent is g.parent`, `g.file is g.file` and `g['subgroup'] is ['subgroup']` all return false. This seems to be a good way to design things: you only hold on to a reference to some root group controller. All objects are created dynamically. Then you have a single reference to destroy when you want to close the whole thing down.

jrs65:
That does sound like it would get rid of some complexities.

I think that works in h5py because underlying it all there is a reference to the HDF5 structures, and so when you call `g.file` it can use the underlying C structures to create a new `h5py.File` object. For `memh5` we don't have that, so we would need to keep a valid reference to the root group to implement this.

jrs65:
I think this one is actually implemented? @kiyo-masui

Issue #26, opened 2015-12-21
Implement __delitem__ in memh5.


kiyo-masui:
There is no way to delete members in memh5. Implement this.

ljgray:
I believe this was addressed in #33

PR #33, opened 2016-03-17
Version 0.2.0

kiyo-masui:
The main change is to address issue #18 such that both memh5.MemGroup and memh5.Memdisk group create new objects when an entry is referenced.

Also a `tod` model was created, bring much of CHIME's data format functionality into caput.

Many documentation changed including intersphinx referencing to h5py documentation.


kiyo-masui:
This also addresses #26.

jrs65:
Looks pretty good to me! I'm sure it's going to break things though...

PR #35, opened 2016-04-29
Make the package work when there is no MPI

zuoshifan:
Some work to make the package also work when there is no MPI, so it can work either parallel or serial. May not be very useful.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -695,6 +695,8 @@ class MemDatasetCommon(MemDataset):
 
     Attributes
     ----------
+    data
+    local_data
     shape
     dtype
 
@@ -736,6 +738,14 @@ class MemDatasetCommon(MemDataset):
     @property
     def comm(self):
         return None
+
+    @property
+    def data(self):
+        return self._data
+
+    @property
+    def local_data(self):
+        return self._data
 
     @property
     def shape(self):
@@ -785,6 +795,8 @@ class MemDatasetDistributed(MemDataset):
 
     Attributes
     ----------
+    data
+    local_data
     shape
     global_shape
     local_shape
@@ -811,6 +823,14 @@ class MemDatasetDistributed(MemDataset):
         dset._data = data
         return dset
 
+    @property
+    def data(self):
+        return self._data
+
+    @property
+    def local_data(self):
+        return self._data.local_array
+    
     @property
     def shape(self):
         return self.global_shape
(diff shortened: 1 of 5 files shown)


kiyo-masui:
This looks super useful but I'll defer to review to @jrs65 as he makes the most use of the distributed stuff.

kiyo-masui:
@jrs65 Bump!

jrs65:
Thanks for the patches, this looks useful, but it's going to take me a while to go through it.

I agree that `memh5` should work if there is no MPI. That's something we need to patch. However, I'm not convinced we should spend much effort to make it work if you set `distributed=True` at creation time anywhere (or equivalently if you load a distributed file marked with distributed hints). I think it should either: decide there's no MPI, this doesn't make sense, and throw an exception; or switch down to distributed=False, use non-distributed datasets and offer up a warning, which is probably friendlier.

Partly my reluctance is that at somepoint you've got to draw the line. `mpiutil` was designed to avoid you needing `mpi4py` installed for a few simple cases, but if we let it expand much beyond that we'll end up having to wrap the whole `mpi4py` API. And `mpi4py` really isn't very hard to install on most systems....

If you have any examples where it's really necessary to do this though, that would be really helpful @kiyo-masui @zuoshifan

jrs65:
I guess one example is the CHIME pipeline implicitly assumes that many of the datasets are distributed (and are thus MPIArray's internally), such that ignoring `distributed=True` (as suggested above) would break things.

For the CHIME pipeline I'm pretty happy just insisting you have MPI installed though. You aren't going to get far without it.

jrs65:
Shifan. I've finally had time to go over this pull request. There's a lot of useful stuff in there so I'd like to merge it pretty much as is.

However, one change I'd like to make is to get rid of the SelfWrapper in `mpiutil`. I'd rather be much more explicit about what's being pulled into `mpiutil`. It doesn't seem to be really needed at the moment, so it would be easy for me to code around the few places it is used. Do you have any other code the really needs it, that would get broken if I removed it?

Also, in the future would it be possible for you could submit pull requests in feature branches? It would be much easier for me to review them. This one has grown by quite a few nominally unrelated commits since the pull request was submitted.

zuoshifan:
Thank you for your appreciation. I use ```SelfWrapper``` mainly want to expose some constants defined in mpi4py.MPI, such as SUM, MAX, etc. for convenience. See https://github.com/zuoshifan/caput/blob/zuo/develop/caput/mpiarray.py line 368 for a use case.

Sorry for the inconvenience to you for many unrelated commits.

jrs65:
I've just merged it as is. If I ever find time, I might go back and reduce the scope of SelfWrapper somewhat. Thanks for all your work!

PR #37, opened 2016-05-10
Ported distributed functionality to tod.concatenate.

kiyo-masui:


jrs65:
Looks good to me. I've just briefly tested it on scinet.

PR #48, opened 2018-02-10
Improvements to how skyfield data is treated

jrs65:
Caput will now initially try and install the skyfield when the package installs. It will also now use attempt to update the data when it is first used and otherwise use existing data.


jrs65:
I also made skyfield a required dependency because that's also been a source of problems for us.

PR #50, opened 2018-05-07
Added support for Cedar

ntdenman:
- New mandatory yaml parameter: `cluster' can be `gpc' or `cedar' 
- New optional parameters: `mem' defaults to `128000M', `account' defaults to `rpp-krs'
- `ppn' now has two defaults, 8 on gpc and 32 on cedar

### DIFF
--- a/scripts/caput-pipeline
+++ b/scripts/caput-pipeline
@@ -70,31 +70,31 @@ def queue(configfile, nosubmit=False):
         raise Exception('Job time is required.')
     clusterconf['time'] = conf['time']
 
-    if 'cluster' not in conf:
-        raise Exception('Cluster is required')
-    elif (conf['cluster'] == 'gpc'):
-        is_gpc = True
-        is_cedar = False
-    elif (conf['cluster'] == 'cedar'):
-        is_cedar = True
-        is_gpc = False
+    # Set queueing system w. per-cluster defaults
+    cluster_defaults = {
+        'gpc' : {'ppn': 8, 'mem': '16000M', 'queue_sys': 'pbs', 'account':None,},
+        'cedar': {'ppn': 32, 'mem': '128000M', 'queue_sys': 'slurm', 'account':'rpp-krs',}
+    }
+    if 'cluster' in conf:
+        clusterconf['queue_sys'] = conf['queue_sys'] if 'queue_sys' in conf else cluster_defaults[conf['cluster']]['queue_sys']
+        clusterconf['ppn'] = conf['ppn'] if 'ppn' in conf else cluster_defaults[conf['cluster']]['ppn']
+        clusterconf['mem'] = conf['mem'] if 'mem' in conf else cluster_defaults[conf['cluster']]['mem']
+        clusterconf['account'] = conf['account'] if 'account' in conf else cluster_defaults[conf['cluster']]['account']
     else:
-        raise Exception('Cluster name not recognized')
+        if 'queue_sys' in conf:
+            clusterconf['queue_sys'] = conf['queue_sys']
+        else:
+            clusterconf['queue_sys'] = 'pbs'
+            raise Warning('Queueing system not set, defaulting to PBS')
+        clusterconf['ppn'] = conf['ppn'] if 'ppn' in conf else 8
+        clusterconf['mem'] = conf['mem'] if 'mem' in conf else 16000M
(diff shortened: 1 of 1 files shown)


jrs65:
Nolan. This is looking great. I've added a few comments inline. Mostly about trying to make this more generic.

scripts/caput-pipeline
```diff
@@ -70,13 +70,32 @@ def queue(configfile, nosubmit=False):
         raise Exception('Job time is required.')
     clusterconf['time'] = conf['time']
 
+    if 'cluster' not in conf:
```
jrs65: Parameter shouldn't be mandatory as this will break all previous YAML files. I'd suggest pick a default, and maybe emit a warning if it wasn't set. Alternatively (and probably preferably), you could check the system for the presence of `sbatch` or `qsub` to decide whether it's a PBS or SLURM system.

scripts/caput-pipeline
```diff
@@ -70,13 +70,32 @@ def queue(configfile, nosubmit=False):
...
+        raise Exception('Cluster is required')
+    elif (conf['cluster'] == 'gpc'):
```
jrs65: Options should be `slurm` or `pbs` I would suggest. Where possible this repo should not hardcode CHIME specifics (though I think it's okay to set defaults that make *our* lives a little easier).

cahofer:
This is great! Thanks!

scripts/caput-pipeline
```diff
@@ -70,13 +70,32 @@ def queue(configfile, nosubmit=False):
...
     clusterconf['queue'] = conf['queue'] if 'queue' in conf else 'batch'
     clusterconf['pernode'] = conf['pernode'] if 'pernode' in conf else 1
-    clusterconf['ppn'] = conf['ppn'] if 'ppn' in conf else 8
     clusterconf['name'] = conf['name'] if 'name' in conf else 'job'
 
+    # Cedar-specific optional parameters
+    clusterconf['mem'] = conf['mem'] if 'mem' in conf else '128000M'
+    clusterconf['account'] = conf['account'] if 'account' in conf else 'rpp-krs'
+
+    if is_gpc:
```
jrs65: Actually, maybe a revised version of my previous comment is that there should be four options for `cluster`. `slurm` and `pbs` which just determine the submission template, but leave the default requirements (e.g. ppn and memory at the low values), and then `gpc` and `cedar` which set custom values for the `ppn` and `memory` defaults.

If you're feeling particularly energetic you could factor the last two out into a dictionary which hosts the default parameters. e.g.

```python
cluster_defaults = {
    'gpc' : {
        'ppn': 8, 'mem': '16000M', 'queue': 'pbs',
    },
    'cedar': {
        'ppn': 32, 'mem': '128000M', 'queue': 'slurm'
    }
}

# ...
# Lookup params
clusterconf['ppn'] = conf['ppn'] if 'ppn' in conf else cluster_defaults[cluster]['ppn']
```
then it's easy for us to extend this to new clusters

scripts/caput-pipeline
```diff
@@ -121,5 +158,5 @@ mpirun -np %(mpiproc)i -npernode %(pernode)i -bind-to none python %(scriptpath)s
         os.system('cd %s; qsub jobscript.sh' % pbsdir)
 
 
-if __name__ == "__main__":
```
jrs65: You really don't like double quotes, eh?

ntdenman:
Thanks, Richard - I'm not seeing source comments, are they via the web interface?

jrs65:
They should be there now.

Richard


On 7 May 2018 at 12:01:12, Nolan Thomas Denman (notifications@github.com)
wrote:

Thanks, Richard - I'm not seeing source comments, are they via the web
interface?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/radiocosmology/caput/pull/50#issuecomment-387168386>,
or mute the thread

scripts/caput-pipeline
ntdenman: It's the inconsistency which was getting to me.

jrs65: Yeah. I can get behind that. I must have written this code in the dim and distant past when I was a double quotes kind of guy. I switched at somepoint, but now I'm writing a lot more C++ it's getting confusing again! :)

jrs65:
Thanks for fixing all this up @ntdenman

PR #54, opened 2018-12-23
Python 3 support

jrs65:


jrs65:
Hey @kiyo-masui @ssiegelx let me know what you think of the ported code. For some stuff I may be able to add to the automatic porting code to improve stuff.

jrs65:
Build failed, but it's from a timeout trying to fetch ephemeris data from a NASA FTP site. I think Travis might have been temporarily blocked because the tests have attempted to download the files so many times.

kiyo-masui:
Okay builds worked this time, so you are probably right about being temporarily blocked. Thanks for porting this, and feel free to merge.

jrs65:
@kiyo-masui I've added a few more changes that were needed to make ch_util work. Unfortunately there's some issues with how h5py treats ascii strings in Python 3 (returning them as byte strings) which cause problems (description [here](https://github.com/jrs65/py3port#h5py)). I've tried to work around it by attempting to convert all attributes to unicode when they are copied out an h5py object. Let me know what you think of this.

kiyo-masui:
This failed because it couldn't download the ephemeris data again.

I'll retrigger the build test.

jrs65:
Thanks. I think it's because the NASA ftp site is locking us out for too much usage. Might be able to self host the data it needs and force the tests to use alternative locations.

jrs65:
I'm going to merge this now. I'm not 100% clear if the FTP issue has gone away, but we'll find out.

PR #55, opened 2019-01-08
Settable pipeline loglevel

kiyo-masui:


jrs65:
Looks good. I usually set the logging using the `draco.core.task.SetMPILogging` but that only works for the things derived from `MPITask`.

scripts/caput-pipeline
```diff
@@ -13,8 +13,12 @@ products = None
 
 
 @arg('configfile', help='Configuration file to run.')
-def run(configfile):
+@arg('--loglevel', help='Logging level (DEBUG, INFO, WARNING, ERROR, CRITICAL)')
+def run(configfile, loglevel='INFO'):
     from caput.pipeline import Manager
+    import logging
+    level = getattr(logging, loglevel.upper())
+    logging.basicConfig(level=logging.INFO)
```
jrs65: Doesn't this just always set the log level to `INFO`?

kiyo-masui: I must have been drunk when I wrote this. I'll fix it.

kiyo-masui:
Okay, how's this?

PR #56, opened 2019-01-18
config key exceptions

wulfda02:
Modify config.Reader to raise exceptions if there are discrepancies between keys in the configuration dictionary and Properties

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -138,7 +138,7 @@ class Property(object):
         if self.key in config:
             val = self.proptype(config[self.key])
             obj.__dict__[self.propname] = val
-      
+
     def _set_propname(self, obj):
         # As this config.Property instance lives on the class it's in, it
         # doesn't actually know what it's name is. We need to search the class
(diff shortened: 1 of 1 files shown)


caput/config.py
```diff
@@ -138,7 +138,7 @@ def _from_config(self, obj, config):
         if self.key in config:
             val = self.proptype(config[self.key])
             obj.__dict__[self.propname] = val
-
+
```
jrs65: Extra spaces on line.

caput/config.py
```diff
@@ -173,21 +173,39 @@ def from_config(cls, config, *args, **kwargs):
...
             for propname, clsprop in basecls.__dict__.items():
                 if isinstance(clsprop, Property):
                     clsprop._from_config(self, config)
-
+                    prop_keys.append(clsprop.key)             
+
+        if compare_keys:
+            if set(config_keys)-set(prop_keys):
+                raise Exception("Configuration keys [%s] do not have corresponding properties" 
+                                % ", ".join(set(config_keys)-set(prop_keys)))
```
jrs65: Please put spaces around the `-` for PEP8.

jrs65:
@wulfda02 this looks really great. Thanks for sending in some nice code. I've made a few slight comments. They're trivial as you can see. I'd make them myself, but the PR is in your own fork so I can't do that. I'll add you as a radio cosmology dev so you can commit straight into the repo.

PR #59, opened 2019-01-24
Required property

ketiltrout:
This is a complement to Dallas's enhancement which allows declaring `Property`s which are required to be initialised from the dictionary and have no sensible default value.

It adds an fourth, optional parameter, `required`, to the `Property` constructor which defaults to `False`. If set to `True`, then `AttributeError` will be raised if an attempt is made to access the instanced attribute before it is set.

Also, because setting a default doesn't make sense for these "required" properties (since it can never be returned), `ValueError` will be raised if the `Property` constructor is called with `default` not `None` and `required=True`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ nosetests.xml
 .mr.developer.cfg
 .project
 .pydevproject
+
+# vim backups and swap files
+.*.swp
+*~
--- a/caput/config.py
+++ b/caput/config.py
@@ -71,7 +71,7 @@ class Property(object):
     """Custom property descriptor that can load values from a given dict.
     """
 
-    def __init__(self, default=None, proptype=None, key=None):
+    def __init__(self, default=None, proptype=None, key=None, required=False):
         """Make a new property type.
 
         Parameters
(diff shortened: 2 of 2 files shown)


kiyo-masui:
@ketiltrout @jrs65 Could we just require that it be set if the default is None?

jrs65:
I agree with @kiyo-masui here. Why not just make that happen if the default is `None`? That's what default None was intended to be for, although that probably isn't made explicit anywhere.

ketiltrout:
While it is true that the default isn't needed if `required=True`, if we make `default=None` act as a "required" parameter and raise an exception if no value is provided, then we can't have optional parameters without forcing them to have a default value.

Say, I have an optional numeric parameter which can take any value, but doesn't have a sensible default.  In that case, I need a non-numeric object to use a the default value so the code can distinguish not-set from any valid value.  That's what the `None` default should be used for, not as a flag to determine the properties of the property

jrs65:
Okay. I think I can start to see what you are saying here, but it still seems a little redundant. Is there an actual example where this has cropped up that we need to fix this for.

As an alternative that I think has the semantics that we all agree on, we could just move the `default` (and `key`) arguments into a `**kwargs` dict and then we can tell if `default` was actually set or not? i.e. if `default` was set the parameter was optional, otherwise it is required.

ketiltrout:
Do you mean an example of an optional parameter with no default?  No, I don't have one of those.  But parameters which are required would be helpful for that `dias` issue mentioned above.

I do like your alternate `kwargs` proposal better than what I have here, though. I'll see if I can implement something we can take a look at.

jrs65:
Yeah. I'm just wondering if we're engineering around a problem that doesn't actually exist. We already do have required parameters, it's just the ambiguity of optional parameters with no meaningful default of the same type we don't support.

Let's go for the kwargs solution if we do anything.

PR #60, opened 2019-02-01
Fix cyclic references

kiyo-masui:
Here are two changes that fix cyclic references and garbage collection.

Changing `MemDataset._storage_root` to a weak reference fixes what I think is the last remaining cyclic reference. I found this using the excellent `objgraph` package. It is as simple as `import objgraph; objgraph.show_refs(data, filename='sample-graph.png')`.

The second is less backward compatible: it removes automatic file closure. This is more inline with how h5py closes files. More importantly, @rmck1 found the following in the python docs: "Objects that have del() methods and are part of a reference cycle cause the entire reference cycle to be uncollectable.", so the existence of __del__ makes the whole container less robust to memory leaks. Seems like a good idea to get rid of it.

Thanks @rmck1 for figuring most of this out!

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -75,6 +75,7 @@ import collections
 import warnings
 import posixpath
 from ast import literal_eval
+import weakref
 
 import numpy as np
 import h5py
@@ -630,7 +631,7 @@ class MemGroup(_BaseGroup):
 
         # Set the properties of the new dataset
         new_dataset._name = posixpath.join(parent_name, name)
-        new_dataset._storage_root = self._storage_root
+        new_dataset._storage_root = weakref.ref(self._storage_root)
         return new_dataset
 
     def dataset_common_to_distributed(self, name, distributed_axis=0):
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -583,6 +583,10 @@ class MemGroup(_BaseGroup):
         if not self.distributed and distributed:
             raise RuntimeError('Cannot create a distributed dataset in a non-distributed group.')
 
+        # Set the properties of the new dataset
+        full_path_name = posixpath.join(parent_name, name)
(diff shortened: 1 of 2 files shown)


jrs65:
Ah great. Yeah `objgraph` is what I used last time I had to figure this one out!

jrs65:
This looks good. If we're getting rid of the auto-closing, can we at least add context manager support so people can do:

```python
with memh5.BasicCont.from_file('testfile.h5', ondisk=True) as f:
    # <do stuff>
    pass
```
 I think it's as simple as adding `__enter__` and `__exit__` methods like exist in `h5py` [here](https://github.com/h5py/h5py/blob/c5859441d47e58be185daf8dc812bd328a8867bb/h5py/_hl/files.py#L435). I reckon it would be 10 lines max. That way people can avoid leaking files, and keep working even more in line with what they expect from `h5py`.

Rest of the changes seem great. Thanks!

kiyo-masui:
You're right: the problem is that the storage holds MemDatasets directly, not some underlying version of them. This contrasts MemGroups, which aren't stored (Storage objects are instead, which are converted to MemGroups)

I guess we need to abstract MemDatasets? will think about it later

jrs65:
> I guess we need to abstract MemDatasets? will think about it later

An easier version might be to: 

- Keep storing datasets internally with the container as MemDatasets. Don't set the _storage_root.
- When a user makes a call that returns a dataset, make a *shallow* copy of the internal dataset, set the `_storage_root` and then return the new clone.

I think that would, remove the reference cycles, keep a strong reference to the root on user facing datasets, and not require any new abstractions.

kiyo-masui:
Here is candidate new code. I still need to check it for cyclic references, but I think they should be fixed.

kiyo-masui:
@jrs65 Looks good from a cyclic reference point of view.

PR #62, opened 2019-02-11
Concat dist bugfix

ssiegelx:
This fixes two bugs.  The first is just a `type() is list` comparison that no longer works in newer versions of python.

The second occurs when concatenating distributed datasets that I think may have been introduced in [4343059](https://github.com/radiocosmology/caput/commit/434305988e4e4260be19320080f596efe22e8204#diff-b295d0d06e0d6df9609f55b50997befc).

Applying the time slice to `dataset` converts it from an instance of `MemDataset` to either an `MPIArray` or `np.ndarray`.

For distributed datasets, this causes the `isinstance(dataset, memh5.MemDatasetDistributed)` condition to fail, and the output dataset is initialized as non-distributed.

The proposed fix converts back to either `MemDatasetCommon` or `MemDatasetDistributed` after applying the time slice.  It determines which one to use by checking if `dataset` is an instance of `MPIArray`.  People that are more familiar with the code may have a better way to do this.

### DIFF
--- a/caput/pipeline.py
+++ b/caput/pipeline.py
@@ -1296,11 +1296,12 @@ def _format_product_keys(spec, name):
         # Default to empty if not present.
         return []
     # Turn into a sequence if only one key was provided.
-    if not type(out_prod) is list:
+    if not isinstance(out_prod, list):
         # Making this a tuple instead of a list is significant.  It only
         # impacts the 'out' product key and affects how return values are
         # unpacked.
         out_prod = (out_prod,)
+
     # Check that all the keys provided are strings.
     for prod in out_prod:
         if not isinstance(prod, basestring):
--- a/caput/tod.py
+++ b/caput/tod.py
@@ -43,6 +43,7 @@ import numpy as np
 import h5py
 
 from . import memh5
+from . import mpiarray
 
 
 class TOData(memh5.BasicCont):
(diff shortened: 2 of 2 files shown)


jrs65:
Sorry. Should be able to look over the code tomorrow.

kiyo-masui:
This is inconsistent with our goal of having `memh5` emulate `h5py` as closely as possible. Slicing an `h5py` dataset returns a numpy array, so I think the original behaviour of is what we wanted.  Can the issue you are having be fixed downstream with something like `isinstance(dataset, memh5.MemDatasetDistributed) or sinstance(dataset, MPIArray)`?

jrs65:
@kiyo-masui isn't that what this change is actually doing, in that it's actually fixing down stream? It's not a change to `memh5` itself, it's a change such that when we concatenate TOD objects the datasets come out right.

kiyo-masui:
Ah yes, I misinterpreted the comments. I should have looked at the source!

jrs65:
Looks good to me too. I'll merge now. Thanks for tracking this down! Was this the source of the issues you were having with distributed datasets @ssiegelx ?

ssiegelx:
@jrs65 Yup this fixed the problem I was having.

jrs65:
Okay great. Rick was looking into a rewrite of the redistribution code, but it sounds like we can lower the priority of that then. We should still do it in the near future, but it's no longer urgent.

Issue #43, opened 2017-02-18
Write a proper redistribute for MPIArray


jrs65:
At the moment the redistribute routine uses an old routine I wrote `transpose_blocks` that was designed for doing a block wise transpose of a 2D(-ish) array. It spends a lot of time massaging things to be able to feed into this somewhat restricted routine and would be far better if we just implemented the redistribution itself.

nritsche:
closed by #63

PR #63, opened 2019-02-14
rewrite MPIArray redistribute without transpose_blocks

nritsche:
Instead of receiving blocks in-place, this creates a numpy array for each block, that is then copied into the new local array. This is because slicing by indexes for the local array on a variable axis does not result in a contiguous array accepted by `mpi4py.MPI.Recv`. Suggestions are welcome!
#43

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -420,44 +420,108 @@ class MPIArray(np.ndarray):
         try:
             mpiutil.typemap(self.dtype)
         except KeyError:
-            if self.comm is None or self.comm.rank == 0:
+            if self.comm.rank == 0:
                 import warnings
-                warnings.warn('Cannot redistribute array of compound datatypes. Sorry!!')
+                warnings.warn('Cannot redistribute array of compound datatypes.'
+                              ' Sorry!!')
             return self
 
-        # Construct the list of the axes to swap around
-        axlist_f = list(range(len(self.shape)))
+        # Get a view of the array
+        arr = self.view(np.ndarray)
 
-        # Remove the axes we are going to swap around
-        axlist_f.remove(self.axis)
-        axlist_f.remove(axis)
+        if self.comm.size == 1:
+            # only one process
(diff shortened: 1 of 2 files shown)


caput/mpiarray.py
```diff
@@ -420,44 +420,108 @@ def redistribute(self, axis):
...
+
+            # For each node iterate over all sends and wait until completion
+            for ir, ic, request in requests_send:
+
+                stat = mpiutil.MPI.Status()
+
+                request.Wait(status=stat)
+
+                if stat.error != mpiutil.MPI.SUCCESS:
+                    print
```
jrs65: In Python 3 `print` is now a function (and in python 2 with the `__future__` imports. Why this doesn't break immediately with a `SyntaxError` is quite fun (it's the the fact that the "argument" is on the next line that means it can pass).

caput/mpiarray.py
```diff
@@ -420,44 +420,108 @@ def redistribute(self, axis):
...
+            # For each frequency iterate over all receives and wait until
+            # completion
+            for ir, ic, request in requests_recv:
+
+                stat = mpiutil.MPI.Status()
+
+                request.Wait(status=stat)
+
+                if stat.error != mpiutil.MPI.SUCCESS:
+                    print
```
jrs65: Print statement needs fixing here as well.

caput/tests/test_mpiarray.py
```diff
@@ -20,7 +20,7 @@
 
 
 
-class TestMPIAray(unittest.TestCase):
+class TestMPIArray(unittest.TestCase):
```
jrs65: Presumably this covers redistribution? Too lazy to look.

caput/mpiarray.py
nritsche: It does.

caput/tests/test_mpiarray.py
nritsche: It does

PR #66, opened 2019-03-02
Parallel container IO using h5py and MPI

jrs65:
Uses HDF5/h5py parallel features to allow concurrent writing of datasets and
more efficient reading.

I have noticed some slight issues using h5py collective parallel routines in
test examples when using OMPIO is the MPI-IO implementation (ROMIO did not seem
to have this problem). At the moment I'm still investigating. When cedar comes
back I'll try to test whether this is a reasonable concern.


jrs65:
There doesn't seem to be an issue running on cedar with this.

jrs65:
Thanks Rick!

PR #68, opened 2019-03-22
Memh5 bugfixes

jrs65:


jrs65:
@kiyo-masui tagged you here to have a look at this one.

The issue is one that came out of your recent h5 changes. Because everything now is given as a view, simply calling redistribute on each view just redistributes the view. You have to operate on the underlying storage. It might be better for the redistribute method on each dataset to do this, though I'm unclear.

PR #69, opened 2019-03-25
[WIP] moving weighted average

nritsche:
Please review the implied function definition.
I noticed that maybe I am falsely assuming that the number of weights equal the size of the moving window, not the length of the values.


jrs65:
Mostly looks good, but the weights should be the same length as the data (as you already surmised). I think that means you'll need to pass in an extra length parameter.

nritsche:
#73

nritsche:
Without parallelization:
caput/tests/test_moving_weighted_median.py 1D moving weighted median with 10000 elements took 0.64s with a simple python implementation vs 0.0075s in cython.

nritsche:
@jrs65 Using a split median, the result of a weighted median can be ambiguous if the weights can be 0:

Values `2,2,5,5,9,9` with weights `0,8,0,5,0,3` can have a weighted median of `mean(2+5+5)` or `mean(2+5)`, depending on how the 2 values `5` with different weights are sorted.

nritsche:
TODO:
- [ ] OpenMP
- [ ] option to select split median / upper / lower weighted median
- [ ] N-dimensional implementation

nritsche:
I think this is finally ready.
 16 * 20,000 elements and a window of size 51*51 takes 1.2s now.
1000 * 1000 elements with the same window takes 15s.

jrs65:
That's great! Is that running in more than 1 thread?

For reference `scipy.signal.medfilt2d` took 21s for the same size (it doesn't do the weighting though).

```
In [12]: a = np.random.random((1000, 1000))

In [13]: %timeit -n 1 -r 1 b = ss.medfilt2d(a, (51, 51))
21.4 s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

nritsche:
Yes, that's 8 threads on my laptop.

Profiling it showed it's keeping the tree balanced that takes most of the time. Biggest speedup I found so far was the time it needed to find the shared_ptr's when moving upwards through the tree.

ssiegelx:
Hey @nritsche, we are finally getting around to testing the `moving_weighted_median`.   We've confirmed that it yields the same results as the moving nan median that we were using previously for a wide variety of 2D data and weight matrices (while being at least a factor of 10 times faster and using significantly less memory).

However we've encountered a few scenarios where it results in a segmentation fault.  It appears to be related to having a region of the weight matrix larger than the window that is all zeros.  If this region is up against index 0 of the 1st dimension of the matrix, then there isn't a problem and the output is as expected.  However, if this region is not up against index 0 of the 1st dimension, then there will be a segmentation fault.

[I've created a test script that can be downloaded here](https://drive.google.com/file/d/1OYP6f--MTkQLxXpmpvCzk5oGJ5vLFCnN/view?usp=sharing).  If you run:

```python test_moving_weighted_median.py central```

it should reproduce this problem.  If you replace `central` with any of the following test cases `small_edge1`, `edge1`,  `large_central`,  `large_edge1`   they will also crash.

The following test cases pass:  `small_central`, `small_corner`, `small_edge0`, `corner`, `edge0`, `large_corner`, `large_edge0`.

 I don't think I will have time to dig into the code, so I wanted to make you aware of this issue.

PR #70, opened 2019-04-02
Allow read to be distributed along specified axis

jrs65:
This is useful as it may allow you to use a more appropriate axis for the read, which can be useful when the dataset is chunked.

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -544,7 +544,7 @@ class MPIArray(np.ndarray):
         return enumerate(range(start, end))
 
     @classmethod
-    def from_hdf5(cls, f, dataset, comm=None, axis=0):
+    def from_hdf5(cls, f, dataset, comm=None, axis=0, sel=None):
         """Read MPIArray from an HDF5 dataset in parallel.
 
         Parameters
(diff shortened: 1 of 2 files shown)


caput/mpiarray.py
```diff
@@ -565,18 +565,28 @@ def from_hdf5(cls, f, dataset, comm=None):
...
         gshape = dset.shape
+        naxis = len(gshape)
         dtype = dset.dtype
-        dist_arr = cls(gshape, axis=0, comm=comm, dtype=dtype)
 
-        start = dist_arr.local_offset[0]
-        end = start + dist_arr.local_shape[0]
+        # Check that the axis is valid and wrap to an actual position
+        if axis < -naxis or axis >= naxis:
+            raise ValueError("Distributed axis not in range (%i, %i)" % (-naxis, naxis-1))
```
nritsche:
```suggestion
            raise ValueError("Distributed axis %i not in range (%i, %i)" % (axis, -naxis, naxis-1))
```

caput/mpiarray.py
```diff
@@ -544,7 +544,7 @@ def enumerate(self, axis):
         return enumerate(range(start, end))
 
     @classmethod
-    def from_hdf5(cls, f, dataset, comm=None):
+    def from_hdf5(cls, f, dataset, comm=None, axis=0):
         """Read MPIArray from an HDF5 dataset in parallel.
 
         Parameters
```
nritsche: add `axis` to docstring

jrs65:
I added a little bit more functionality here. Sorry! I think that's the last part I actually need.

PR #71, opened 2019-04-10
Context manager for lockfiles

jrs65:
A context manager for wrapping around file creation operations that
automates the creation and cleaning of lock files, as well as removal of
the temporary file when there is a failure.


caput/misc.py
```diff
@@ -16,6 +16,7 @@
 from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 # === End Python 2/3 compatibility
 
+import os
```
kiyo-masui: I think `shutil` is the preferred package to use, not that it matters.

PR #75, opened 2019-06-13
fix: partition parallel IO to be under the 2GB HDF5 limit

jrs65:
This includes a large refactoring of the mpiarray IO code to break it up
and add the ability to split IO into a chunks to ensure that they fit
within HDF5's limits. Additionally collective IO is turned off for
writes along axis=0, and parallel IO entirely for reads along axis=0 to
improve IO performance.

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -110,9 +110,14 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 from past.builtins import basestring
 import os
 import time
+import logging
+
 import numpy as np
 
 from caput import mpiutil, misc
+
+
+logger = logging.getLogger(__name__)
 
 
 class _global_resolver(object):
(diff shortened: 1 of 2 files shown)


caput/mpiarray.py
```diff
@@ -807,3 +840,107 @@ def _to_hdf5_serial(self, filename, dataset, create=False):
...
+        partitions : list of slice objects
+            List of slices.
+        """
+        from mpi4py import MPI
+
+        threshold_bytes = threshold * 2**30
+        largest_size = self.comm.allreduce(self.nbytes, op=MPI.MAX)
+        num_split = int(np.ceil(largest_size / threshold_bytes))
+
+        if skip or self.ndim == 1:
```
tristpinsm: should this be `if skip or num_split == 1:` ? Seems like if you are not splitting up the IO you can just return at this point. What about cases where the array is 1D but greater than 2GB?

caput/mpiarray.py
```diff
@@ -807,3 +840,107 @@ def _to_hdf5_serial(self, filename, dataset, create=False):
...
+        # Try and find the axis to split over
+        for split_axis in range(self.ndim):
+            if split_axis != self.axis and self.global_shape[split_axis] >= num_split:
+                break
+        else:
+            raise RuntimeError(
+                "Can't identify an IO partition less than %.2f GB in size: "
+                "shape=%s, distributed axis=%i" % (threshold, self.global_shape, self.axis)
+            )
+        if num_split > 1:
```
tristpinsm: If my comment above is correct, this should always be `True`

caput/mpiarray.py
jrs65: The issue here is that it can only split across a different axis to the distributed axis, so if there's only one axis it can't do anything. Also, if `num_split == 1` that's a totally normal circumstance to be in (it just means the array is small enough it doesn't need to be split). You are right though that if num_split does equal one then we could exit at this point. I can try and change the logic around to make that clear, also to warn if there is only axis which is too large.

tristpinsm: Yeah, that makes sense. Does it cause an mpi-io issue if there is only one axis and it is over 2GB? That should be a pretty rare case for us, but maybe a warning could help track that down if it ever comes up.

PR #77, opened 2019-06-24
Chunked HDF5 with MPI

tristpinsm:
I finished up adding support for chunks/compression to `MPIarray`s last week. I tried to mimick how these options are passed in `h5py`.

I've tested this in the pipeline with writing out a `SiderealStream` and it seems to work as expected.

One thing to note is that collective IO is required to do chunked writes, so I've had to mess around a bit with the logic that figures out when to use this.

I also have not done any performance testing. This is obviously going to be important, but I know Richard has been investigating this in previous code updates, so I thought I would start the pull request and get some input on how to test it, what to compare it too.

I'm going to open a separate PR in `draco` with updates to the containers that make use of chunking/compression.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -533,6 +533,9 @@ class MemGroup(_BaseGroup):
         data=None,
         distributed=False,
         distributed_axis=None,
+        chunks=None,
+        compression=None,
+        compression_opts=None,
         **kwargs
     ):
         """Create a new dataset.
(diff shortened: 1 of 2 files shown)


tristpinsm:
The `draco` PR is https://github.com/radiocosmology/draco/pull/32

tristpinsm:
OK, looks like the unit tests have revealed an issue I hadn't anticipated... I'll have to spend a bit more time figuring out what to do there.

jrs65:
Ooooh awesome. This is mostly great timing! I'm just starting to get worried about the volume of data in the daily processing (600 GB/day) and this would probably be a great saver.

tristpinsm:
OK I picked this up today and I think I resolved the issues. Caveats about performance remain, but at least it seems to work.
@jrs65

PR #88, opened 2019-09-07
build: Enforce numpy>=1.16 everywhere

### LINKED ISSUE #87: Fix numpy requirement version bug
When installing caput the following error may occur:

```Could not find a version that satisfies the requirement numpy==1.15.1 (from versions: 1.11.3, 1.12.1, 1.13.0, 1.13.1, 1.13.3, 1.14.0, 1.14.1, 1.14.2, 1.14.3, 1.14.5, 1.15.0, 1.15.2, 1.15.4, 1.16.0, 1.16.2, 1.16.3)
  No matching distribution found for numpy==1.15.1```

I believe this could be solved by changing the requirement in `pyproject.toml` from `'numpy==1.15.1` to `'numpy>=1.15.1` ?

@anjakefala @ketiltrout @nritsche

ketiltrout:
If we really need to maintain three different lists of requirements, let's at least make them all the same.

`numpy-1.16` has a forwards-incompatible ABI change which seems to be causing grief (see numpy/numpy#12785), but if we enforce a NumPy after the change, we should be good, hopefully.

Closes #87

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ['setuptools', 'wheel', 'Cython', 'future', 'numpy==1.15.1']
+requires = ['setuptools', 'wheel', 'Cython', 'future', 'numpy>=1.16']
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-numpy
+numpy>=1.16
 h5py
 PyYAML
 skyfield>=1.0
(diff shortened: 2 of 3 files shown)


ketiltrout:
Hmmm, I could see an argument here for fixing the `pyproject.toml` version to, say, 1.16.0, and leaving the install requirements at `>=1.16`, the worry being that some future `numpy` is going to have another forwards-incompatible ABI break.

As it stands, I've verified that I can build `caput` with NumPy-1.17.2 and then run it using NumPy-1.16.0, so it's not an issue at the moment.  I'm tempted to leave this as-is to prevent more problems like @fandinomat reports where particular versions of packages are not available for particular platforms.

PR #96, opened 2019-12-12
feat(caput-queue): Option for a python script to be run before pipeline.

tristpinsm:
A new section in the pipeline config will be parsed as a python script
and included in the jobscript before the main pipeline.


jrs65:
@tristpinsm as we talked about yesterday, I think we have a cleaner way to do this one, so I'm going to close it without merging. Thanks anyway!

PR #98, opened 2020-01-16
fix(time): include required skyfield data in the repo (via Git LFS)

jrs65:


nritsche:
Should we by default first try to download the skyfield data and only take the cached in case that fails. I understand it gets updated some times.

jrs65:
Indeed, those files get updates every now and again.

That's actually what it does whenever it runs (i.e. whenever you try and use many of the routines in `caput.time`), it tries to see if the data has expired (most of the data has an associated expiration data) and if it has tries to down load the new timescale data. I think that should mostly be sufficient.

I could put it back in (and make it be possible to disable for the Travis tests), but I'll probably leave it out for the moment. The servers it pulls from have been extremely flaky recently (and I think occasionally try and block travis checking them from too many requests), and so if I just keep updating the internal versions every ~6 months we'll be mostly fine.

nritsche:
> I could put it back in (and make it be possible to disable for the Travis tests), but I'll probably leave it out for the moment. The servers it pulls from have been extremely flaky recently (and I think occasionally try and block travis checking them from too many requests), and so if I just keep updating the internal versions every ~6 months we'll be mostly fine.

What I thought of is making skyfield do it's thing, but have the cached files in case that fails. That way it wouldn't be so bad if you forget to update it here in ~6 months.

anjakefala:
(This is an exciting PR...)

jrs65:
> (This is an exciting PR...)

Haha. Yeah. I've probably rebased and force pushed about 10 times so I hope you're not getting all the notifications.

jrs65:
> What I thought of is making skyfield do it's thing, but have the cached files in case that fails. That way it wouldn't be so bad if you forget to update it here in ~6 months.

Yeah, that's what I figured, I just don't want to do it myself :) Happy to accept PRs for it though!

PR #99, opened 2020-01-22
refactor: remove internal skyfield data and download cache during testing

jrs65:
This reverts a misguided attempt to keep the skyfield data in Github
LFS, which had issues because of the absurdly low bandwidth quota.
Instead it removes all the special handling of skyfield data except
during testing where it will download a copy from a CHIME server.


caput/tests/test_time.py
```diff
@@ -16,6 +16,23 @@
 from caput import time as ctime
 
 
+def setUpModule():
```
nritsche: Is this the only test that triggers the download? Maybe this function should be on a higher level as a helper available to tests.

jrs65: It originally was, but then I figured this was the only place that needed it, so just put it directly in here. I think if we ever place tests for the skyfield code in another module we'll need to do that, but I don't really see that happening.

PR #101, opened 2020-01-29
fix(runner): Add call to group at end of file.

fandinomat:
It seems runner is missing the call to group at the end of file?

### DIFF
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -299,5 +299,6 @@ def fixpath(path):
     """Turn path to an absolute path"""
     return realpath(abspath(path))
 
+# This is needed because the queue script calls this file directly.
 if __name__ == '__main__':
     cli()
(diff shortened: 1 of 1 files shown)


jrs65:
It shouldn't need it I think, the `setup.py` script will install the hooks that call `cli()`. Were you having some problem that suggests it should?

fandinomat:
Yes, I was getting no response when scheduling a job.
I will try a fresh install and report if I still get a problem.

fandinomat:
OK, I realize what created my problem in the first place. For the setup to call cli() the script must be ran as `caput-pipeline', not a direct call to runner.py.
The problem is, when the command `caput-pipeline queue' is issued, the script created calls runner.py directly and then nothing happens because runner.py never calls cli().

jrs65:
Ah. Thanks Mateus, that makes much more sense. I can see why you needed that fix now. I actually think I preferred the first fix you made (i.e. adding the `if __name__ == ...` block) so can you revert to that commit and add a comment explaining why it's needed.

fandinomat:
Done.

fandinomat:
Done.

fandinomat:
@jrs65 Ping.

PR #105, opened 2020-02-18
Fix crash in moving weighted median and a new config property

jrs65:


jrs65:
Added some tests. Merging now.

PR #106, opened 2020-02-19
feat(Manager): add metadata options

### LINKED ISSUE #104: add version strings to Property.global_config in Reader.read_config
get a list of module names from global config level, get these module's `__version__ and add it to task aka `Property` instances

nritsche:
- version strings for list of modules
- config dump

Closes #104

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -80,6 +80,7 @@ import warnings
 import posixpath
 from ast import literal_eval
 
+import json
 import numpy as np
 import h5py
 
@@ -1987,6 +1988,9 @@ def copyattrs(a1, a2):
     # Make sure everything is a copy.
     a1 = attrs2dict(a1)
 
+    # When serializing dictionaries, add this in front of the string
+    json_prefix = "!!_memh5_json:"
+
     def _map_attr(value):
 
         # Any arrays of numpy type unicode strings must be transformed before being copied into HDF5
@@ -2003,10 +2007,21 @@ def copyattrs(a1, a2):
                 return value
 
         # If we are copying into memh5 ensure that any string are unicode
-        return bytes_to_unicode(value)
+        value = bytes_to_unicode(value)
+
+        # Deserialize "special" _json values
+        if not isinstance(value, str):
+            return value
+        if value.startswith(json_prefix):
+            return json.loads(value[len(json_prefix) :])
+        return value
 
     for key in sorted(a1):
-        a2[key] = _map_attr(a1[key])
+        # Serialize dict values to json
+        if isinstance(a1[key], dict):
+            a2[key] = json_prefix + json.dumps(a1[key])
+        else:
+            a2[key] = _map_attr(a1[key])
 
 
 def deep_group_copy(g1, g2):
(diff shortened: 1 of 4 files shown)


caput/pipeline.py
```diff
@@ -312,6 +312,8 @@
 from future.builtins import *  # noqa  pylint: disable=W0401, W0614
 from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 
+from six import string_types
```
jrs65: I think here you should use `from past.builtins import basestring` and use `basestring` where you've used `string_types`. That's what we're doing elsewhere in the code.

caput/pipeline.py
```diff
@@ -405,6 +448,10 @@ class Manager(config.Reader):
     cluster = config.Property(default={}, proptype=dict)
     tasks = config.Property(default=[], proptype=list)
 
+    # Options to be stored in self.all_tasks_params
+    versions = config.Property(default=[], proptype=_get_versions, key="save_versions")
```
jrs65: Can you document these properties in the doc string above?

caput/pipeline.py
```diff
@@ -565,9 +627,12 @@ def _setup_task(self, task_spec):
                     msg = "Parameter group %s not found in config." % param_key
                     raise PipelineConfigError(msg)
 
+        # add global params to params
+        params.update(self.all_tasks_params)
```
jrs65: I think you want the parameter update to be done the other way around. i.e. local params should be used in preference to global ones.

caput/memh5.py
```diff
@@ -2003,10 +2008,16 @@ def _map_attr(value):
...
+
+        # Deserialize "special" _json values
+        if key.endswith("_json"):
+            return json.loads(value)
+        else:
+            return value
 
     for key in sorted(a1):
-        a2[key] = _map_attr(a1[key])
+        a2[key] = _map_attr(key, a1[key])
```
jrs65: Should the keys get renamed in here? i.e. `version_json` -> `version` when being read (and vice-versa),

nritsche: Then it wouldn't get serialized again if it's written to hdf5 later.

jrs65: I think the solution to that is fairly easy, you just do:
```python
if isinstance(a1[key], (dict, list, tuple)):
    a2[key + "_json"] = json.dumps(a1[key])
```

jrs65: The rationale being that `h5py` can't write those types anyway, so it doesn't matter if you do something special with them.

nritsche: We can do that, but only for `dict`'s, since `tuple`'s and `list`'s are converted to numpy arrays in `_map_attr`.

jrs65: Hmmm. Good point. Then maybe we just convert dicts. I think that's mostly fine.

jrs65: It might be cleaner to not change the key names, but have the serialised value start with a magic string. Something like: `!!json:{"an_example_json": "dict"}`

jrs65: That seems less likely to give any potential clashes to me and unexpected behaviour to me.

jrs65: Maybe even: `!!_memh5_json:{"an_example_json": "dict"}`

PR #109, opened 2020-02-25
Memh5 unicode conversions (and misc fixes)

jrs65:


caput/misc.py
```diff
@@ -196,3 +199,77 @@ def tmpfile(self):
...
+    f : function
+        Callable to inspect.
+
+    Returns
+    -------
+    fullargspec : namedtuple
+        Named tuple with various fields. See `inspect.getfullargspec`.
+    """
+
+    import inspect
```
nritsche: Is there a good reason this isn't ontop? https://www.python.org/dev/peps/pep-0008/#imports

caput/memh5.py
```diff
@@ -2400,3 +2566,207 @@ def bytes_to_unicode(s):
...
+    Returns
+    -------
+    dset :
+        The converted array. If no conversion was required, just returns `arr`.
+    """
+    if has_unicode(dset.dtype):
+        raise TypeError(
+            'Can not write dataset "%s" of unicode type into HDF5.' % dset.name
+        )
+    else:
```
nritsche: I guess you can argue about this, but I find that there should really not be an `else` here.

caput/pipeline.py
```diff
@@ -311,21 +311,18 @@
...
 from future.builtins import *  # noqa  pylint: disable=W0401, W0614
 from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
-from past.builtins import basestring
 
 # === End Python 2/3 compatibility
 
-from future import standard_library
-
-standard_library.install_aliases()
+from past.builtins import basestring
```
nritsche: Isn't this only needed for py2/3 compatibility?

PR #111, opened 2020-02-28
feat(memh5): sel_* parameters

### LINKED ISSUE #108: MemGroup.from_hdf5: add option for axis downselection
Add option to supply parameters like `<axis>_sel : numpy index` to downselect data on axis `<axis>` when reading from hdf5 file

nritsche:
To down-select axes when loading from file.

Alse see: https://github.com/radiocosmology/draco/pull/77

Todo:
- [x] document parameter `selections=`
- [x] deep_group_copy: make selection default `slice(None)`

Closes #108


nritsche:
@jrs65 Do you understand why the py2 CI fails? Do we need py2 support here?

caput/memh5.py
```diff
@@ -2152,40 +2187,89 @@ def _map_json(value):
...
 ):
-    """Copy full data tree from one group to another."""
+    """
+    Copy full data tree from one group to another.
+
+    Copies from g1 to g2. An axis downselection can be specified by supplying the
+    parameter 'selections'. If for example 'g1' is
+
+    ```
+    foo:
```
jrs65: Can you write this example in the standard `doctest` format? If you do the example should automatically get run and tested in the CI run.

caput/memh5.py
```diff
@@ -2152,40 +2187,89 @@ def _map_json(value):
...
+        Deep copy from this group.
+    g2 : h5py.Group
+        Deep copy to this group.
+    selections : dict
+        If this is not None, it should have a subset of the same hierarchical structure
+        as g1, but ultimately describe axis selections for group entries as valid
+        numpy indexes.
+    convert_strings : bool, optional
+        Convert string attributes (or lists/arrays of them) to ensure that they are
+        unicode.
```
jrs65:
```suggestion
    convert_attribute_strings : bool, optional
        Convert string attributes (or lists/arrays of them) to ensure that they are
        unicode.
    convert_dataset_strings : bool, optional
        Convert strings within datasets to ensure that they are unicode.
```

caput/memh5.py
```diff
@@ -2152,40 +2187,89 @@ def _map_json(value):
...
+    """
+
+    if selections is None:
```
jrs65: Can't you remove this along with the branch commented on below.

caput/memh5.py
```diff
@@ -2152,40 +2187,89 @@ def _map_json(value):
...
+        selections = slice(None)
 
     copyattrs(g1.attrs, g2.attrs, convert_strings=convert_attribute_strings)
 
     # Sort to ensure consistent insertion order
     for key in sorted(g1):
         entry = g1[key]
+        if isinstance(selections, dict):
+            selection = selections.get(key, slice(None))
+        else:
```
jrs65: This one.

caput/memh5.py
nritsche: You can't pass `None` as a numpy index.

nritsche: I could make the default argument `slice(None)` but I want it to be pretty.

caput/memh5.py
nritsche: Removing this broke the test. We need to iterate through the items, in case `selections` is a dict.
But because `deep_group_copy` also calls itself recursively, it might not be a dict.

nritsche:
Please squash when merging...

caput/tests/test_selection.py
```diff
@@ -0,0 +1,54 @@
...
+dset1 = dset1.reshape((len_axis, len_axis, len_axis))
+
+dset2 = np.arange(len_axis * len_axis)
+dset2 = dset2.reshape((len_axis, len_axis))
+
+freqs = np.arange(len_axis)
+inputs = np.arange(len_axis)
+ra = np.arange(len_axis)
+
+fsel = slice(5)
```
jrs65: The tests should have a test for a sparse slice (e.g. `slice(1, 8, 2)`)

PR #117, opened 2020-03-03
fix(memh5): serialise datetimes with dict attributes without crashing

jrs65:


caput/memh5.py
```diff
@@ -2139,9 +2140,19 @@ def _map_unicode(value):
 
     def _map_json(value):
         # Serialize/deserialize "special" json values
+
+        # Datetimes often appear in the configs (as they are parsed by PyYAML),
+        # so we need to serialise them back to strings
+        def _convert_datetime(v):
+            if isinstance(v, datetime.datetime):
+                return v.isoformat()
+            raise TypeError("Can not JSON serialise object of type %s" % repr(type(v)))
```
nritsche: It looks like you are supposed to let the default default raise the TypeError: 
https://docs.python.org/3/library/json.html#json.JSONEncoder.default
```
# Let the base class default method raise the TypeError
   return json.JSONEncoder.default(self, o)
```

jrs65: It's not clear how to  do that if I'm not creating my own JSONEncoder instance. Thoughts?

nritsche: Hmm. Do you have a test for the code you added?

jrs65: Nothing clean. I've just been running through an example pipeline that fails.

nritsche:
This will always stay a string and never be de-serialized again by memh5 (when reading from hdf5 into a container. Is this what we want?

jrs65:
All looks good. Thanks for adding the test.

nritsche:
I wasn't done here.

jrs65:
Oh oops. Sorry. I guess we should have re-added a WIP. I skimmed the test and it looked complete. I think it unlikely that anyone clones soon, so if you want to force push to master go for it.

nritsche:
https://github.com/radiocosmology/caput/compare/serialise-datetime-json-attributes?expand=1

PR #118, opened 2020-03-04
memh5: let json.JsonEncoder.default raise the TypeError

nritsche:
I think this is nicer, but not sure if you agree.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2153,7 +2153,7 @@ def copyattrs(a1, a2, convert_strings=False):
         if isinstance(value, dict):
             # Save to JSON converting datetimes.
             encoder = DatetimeJSONEncoder()
-            value = json_prefix + json.dumps(value, default=encoder._convert_datetime)
+            value = json_prefix + encoder.encode(value)
         elif isinstance(value, str) and value.startswith(json_prefix):
             # Read from JSON, keep serialised datetimes as strings
             value = json.loads(value[len(json_prefix) :])
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2146,7 +2146,8 @@ def copyattrs(a1, a2, convert_strings=False):
         def _convert_datetime(v):
             if isinstance(v, datetime.datetime):
                 return v.isoformat()
-            raise TypeError("Can not JSON serialise object of type %s" % repr(type(v)))
+            # Let the default method raise the TypeError
+            return json.JSONEncoder.default(json.JSONEncoder, v)
 
         if isinstance(value, dict):
             # Save to JSON converting datetimes.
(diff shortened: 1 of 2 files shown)


caput/memh5.py
```diff
@@ -2146,7 +2146,8 @@ def _map_json(value):
         def _convert_datetime(v):
             if isinstance(v, datetime.datetime):
                 return v.isoformat()
-            raise TypeError("Can not JSON serialise object of type %s" % repr(type(v)))
+            # Let the default method raise the TypeError
+            return json.JSONEncoder.default(json.JSONEncoder, v)
```
jrs65: I'm pretty sure `json.JSONEncode.default` is an instance method, so this only works because the particular implementation of it doesn't use the `self` argument (or doesn't use it to access instance attributes). Seems pretty risky to rely on that implementation not changing.

If we do want to call the base implementation I would just do a very quick subclass of `JSONEncoder` that moves the `_convert_datetime` implementation into there.

caput/memh5.py
```diff
@@ -2143,15 +2143,17 @@ def _map_json(value):
...
+                if isinstance(v, datetime.datetime):
+                    return v.isoformat()
+                # Let the default method raise the TypeError
+                return json.JSONEncoder.default(self, v)
 
         if isinstance(value, dict):
             # Save to JSON converting datetimes.
-            value = json_prefix + json.dumps(value, default=_convert_datetime)
+            encoder = DatetimeJSONEncoder()
+            value = json_prefix + json.dumps(value, default=encoder._convert_datetime)
```
jrs65:
```suggestion
            value = json_prefix + encoder.encode(value)
```

jrs65: I think that works.

caput/memh5.py
```diff
@@ -2143,14 +2143,17 @@ def _map_json(value):
 
         # Datetimes often appear in the configs (as they are parsed by PyYAML),
         # so we need to serialise them back to strings
-        def _convert_datetime(v):
-            if isinstance(v, datetime.datetime):
-                return v.isoformat()
-            raise TypeError("Can not JSON serialise object of type %s" % repr(type(v)))
+        class DatetimeJSONEncoder(json.JSONEncoder):
+            def _convert_datetime(self, v):
```
jrs65:
```suggestion
            def default(self, v):
```

jrs65:
Looks good. I'll let you rebase the commits and then merge.

nritsche:
I think we can just squash merge this:
```
 refactor(memh5): move _convert_datetime into decoder subclass 

- let json.JsonEncoder.default raise the TypeError 
```

jrs65:
Sounds good. Go for it!

PR #120, opened 2020-03-25
fix(tod): fixes bug where `data` variable overwritten

ssiegelx:
Fixes bug in `_copy_non_time_data` where the `data` variable
was being over written if a dataset without a time axis was
encountered.

Bug introduced in commit 249face8d4e0ab38e91507c0724606687d8515c8.

### DIFF
--- a/caput/tod.py
+++ b/caput/tod.py
@@ -634,12 +634,12 @@ def _copy_non_time_data(
             ):
                 to_dataset_names.append(entry.name)
             elif out is not None:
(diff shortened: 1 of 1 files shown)


jrs65:
Ooops looks like I messed that one up. I'm going to try and resurrect the CI build and then I'll merge.

jrs65:
Alright. It all works fine with the new CI tests so I'm going to merge.

PR #121, opened 2020-03-25
Bug fixed on variable name: pro_argspec.keywords to pro_argspec.varkw

danielemichilli:
A property of misc.getfullargspec was misspelled, I renamed following the suggestion of @jrs65


jrs65:
Thanks Daniele. I'm really unclear why the CI build isn't running - it seems that Travis has broken. Given that your fix seems to be very unlikely to introduce any other problems (famous last words), I'm just going to force merge.

PR #123, opened 2020-03-26
Initial actions workflow

jrs65:


jrs65:
@kiyo-masui Travis has been a massive pain recently, and seemed to stop working for caput. I've replaced it with Github Actions which is more reliable, and faster!

kiyo-masui:
How difficult was it?

jrs65:
Pretty easy actually, but I've spent the last 4 days trying to get it going for kotekan, so I'm very familiar with it.

Basically it's pretty similar to the travis format. A bit more verbose, in that you have to explicitly ask for various steps like actually checking out the repo, but more powerful overall. And it seems to be fast, so overall I'm very impressed. I'll probably port many of our CI tests over to it.

PR #129, opened 2020-04-30
feat(pipeline): improve log level config

nritsche:
- Allow to pass a dict to config section 'pipeline/logging' that defines a
log level per module.
- Add option 'logging' to task config that defines the loglevel for the
module the task uses (takes priority over the above).

Fixes https://github.com/chime-experiment/Pipeline/issues/36

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -67,6 +67,12 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 
 # === End Python 2/3 compatibility
 
+import logging
+
+from . import misc
+
+logger = logging.getLogger(__name__)
+
 
 class Property(object):
     """Custom property descriptor that can load values from a given dict.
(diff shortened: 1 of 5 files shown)


caput/config.py
```diff
@@ -397,6 +403,81 @@ def _prop(val):
...
+    --------
+    Should be used like::
+
+        class Project(object):
+
+            loglevels = logging_config({"root": "INFO", "annoying.module": "WARNING"})
+    """
+
+    def _prop(val):
+        config = val
```
jrs65, line 433:
```suggestion
    def _prop(config):
```

caput/config.py
```diff
@@ -397,6 +403,81 @@ def _prop(val):
...
+            raise ValueError(
+                "Expected a string or YAML block for config value 'logging', got "
+                "'{}'.".format(type(config.__name__))
+            )
+
+        # check entries, get module names and warn for duplicates when sorting into new dict
+        checked_config = {}
+        for key, level in config.items():
+            level = level.upper()
+            loglevels = ["DEBUG", "INFO", "WARNING", "ERROR", "NOTSET"]
```
jrs65, line 446: Python really should expose these publicly somewhere. Not much we can do about it though. I would move this declaration out of the loop though.

caput/config.py
```diff
@@ -397,6 +403,81 @@ def _prop(val):
...
+                already_set_to = checked_config.get(module, None)
+                if already_set_to is not None and already_set_to != level:
+                    logger.warning(
+                        "Setting log level for {} to {}, but is already set to "
+                        "{} for module {}. The old value will get ignored.".format(
+                            key, level, already_set_to, module
+                        )
+                    )
+
+                checked_config[module] = level
```
jrs65, line 473: I would recommend removing this check. There's no requirement that logger names have to map to fully qualified Python names (modules, classes etc), and definitely no requirement that they map to modules (although I now see that your function works for both). For one thing I could see us using this mechanism to turn off logging from a specific class somewhere.

Also, although it is certainly not best practice sometimes loading modules can have side effects and so I'm also not sure that importing them in here in a good idea. It would be unexpected at the very least.

Anyway, I'd still recommend removing this check. Although it works within the recommended way of using Python logging, it could break perfectly allowable logging patterns.

caput/config.py
nritsche: you can pass a string to `setLevel` ...from python3.2

caput/config.py
nritsche: meh. I was torn between the 2 options:
- Help people find their typos in the logging config
- Not having to do all this.

I decided for this because it probably gets imported anyways later and because the version number thing does something similar. But I'll remove it

PR #134, opened 2020-05-10
fix(memh5): modify BasicCont.redistribute to prevent reference cycles

jrs65:
The previous implementation caused a reference cycle to be generated preventing the entire container from being automatically freed. The way this happens is quite interesting, at least to me...
- Because of the way python late binds closures, as soon as `.redistribute` is called, the recursive inner function `_tree_crawl` forms a self referencing cycle (via an internal cell object). This reference cycle prevents it being automatically freed when the parent container object's external reference is removed.
- Unfortunately on top of this, the use of `self.comm.rank` binds a reference to `self` into the cyclic closure and means that even when all external references to the container are gone, this cyclic closure still holds one and thus the parent container itself cannot be freed either.

Obviously all of this is fixed by the garbage collector, but we occasionally run out of memory between gc sweeps, so it's better to not generate the reference cycle in the first place.

A minimal fix would just be binding only the value of `rank` into the closure and not `self`, but this patch does the next step and moves to a non-recursive traversal instead.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2038,57 +2038,57 @@ class BasicCont(MemDiskGroup):
         if not isinstance(dist_axis, (list, tuple)):
             dist_axis = [dist_axis]
 
-        # Worker routine to crawl the tree and redistribute any parallel datasets
-        def _tree_crawl(group):
+        stack = list(self._data._storage_root.items())
 
-            for name, item in group.items():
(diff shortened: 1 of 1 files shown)


kiyo-masui:
I imagine this was hard to figure out!

jrs65:
> I imagine this was hard to figure out!

That's an understatement :)

PR #135, opened 2020-06-24
Interferometry pfb routines

jrs65:


jrs65:
@wulfda02 I added a routine to try and convert to XYZ coordinates that should be along the same lines as the one we were discussing before. What do you think?

caput/pfb.py
```diff
@@ -0,0 +1,190 @@
...
+    window : np.ndarray[ntaps * lblock]
+    """
+    # Sampling locations of sinc function
+    X = np.linspace(-ntap / 2, ntap / 2, ntap * lblock, endpoint=False)
+
+    # np.sinc function is sin(pi*x)/pi*x, not sin(x)/x, so we can just X
+    return np.sinc(X)
+
+
+def sinc_hanning(ntap, lblock):
```
kiyo-masui, line 59: Isn't Hanning a misnomer for Hann? I guess we should match whatever the terminology is in CASA...

kiyo-masui: https://en.wikipedia.org/wiki/Hann_function#Name

caput/pfb.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Returns
+    -------
+    window : np.ndarray[ntaps * lblock]
+    """
+
+    return sinc_window(ntap, lblock) * np.hamming(ntap * lblock)
+
+
+def pfb(timestream, nfreq, ntap=4, window=sinc_hamming):
```
kiyo-masui, line 95: I would rename this chime_pfb - pub is too general for a chime-specific code.

caput/pfb.py
```diff
@@ -0,0 +1,190 @@
...
+        # Perform a real FFT (with applied window function)
+        ft = np.fft.rfft(ts_sec * w)
+
+        # Choose every n-th frequency
+        spec[bi] = ft[::ntap]
+
+    return spec
+
+
+def decorrelation_ratio(ntap, lblock, window=sinc_hamming, oversample=4):
```
kiyo-masui, line 140: Are functions mutable? I'm not sure if the best practice is to set the default to be `None` just say this corresponds to sing_hamming in the doctoring.

caput/pfb.py
kiyo-masui: Oh, I think this is just an issue with the doc-string... this isn't chime-specific at all.

caput/interferometry.py
```diff
@@ -0,0 +1,121 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+from future.builtins import *  # noqa  pylint: disable=W0401, W0614
+from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
+
+# === End Python 2/3 compatibility
+
+import numpy as np
+
+
+def sph_to_ground(ha, lat, dec):
```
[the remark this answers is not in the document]
wulfda02: I verified that is equivalent to my routine to within round-off error.  My version uses a rotation matrix, but I like this better since it has fewer lines and better supports broadcasting if `ha` and `dec` have different shapes.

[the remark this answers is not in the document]
wulfda02: I also have the inverse function if that would be useful (I can also make my own pull request):

```
def ground_to_sph(x, y, lat):
    """Get the CIRS coordinates.
    Latitude is given in radians. Assumes z is positive
    Parameter
    ---------
    x : array_like
        The East projection of the angular position
    y : array_like
        The North projection of the angular position
    lat : array_like
        The latitude of the observatory.
    Returns
    -------
    HA, dec: array_like
        Hour Angle and declination in radians
    """

    z = np.sqrt(1 - x**2 - y**2)
    
    xe = z * np.cos(lat) - y * np.sin(lat)
    ye = x
    ze = y * np.cos(lat) + z * np.sin(lat)
    
    HA = -1 * np.arctan2(ye, xe)
    dec = np.arctan2(ze, np.sqrt(xe**2 + ye**2))
    
    return HA, dec
```

jrs65:
Thanks for the comments everyone. My github notifications were broken so I didn't see them at the time.

@wulfda02 I'll just introduce your inverse routine directly in this PR. Thanks for that one!

caput/interferometry.py
```diff
@@ -0,0 +1,121 @@
...
+        The vertical separation on wavelengths (increases to the sky!)
+
+    Returns
+    -------
+    phase : np.ndarray
+        The phase required to *correct* the fringeing. Shape is
+        given by the broadcast of the arguments together.
+    """
+
+    phase = -2.0j * np.pi * project_distance(ha, lat, dec, u, v, w)
```
jrs65, line 119: Ooops. This is a bug. Memo to self.

jrs65:
@kiyo-masui did the refactor you suggested. Definitely nicer now. Thanks!

jrs65:
I think this is good now, so I'm merging.

PR #137, opened 2020-08-24
fix(mpiarray): bug when selecting data during distributed read

ssiegelx:
Currently MPIArray.from_hdf5 uses the shape of the post-selection dataset
to form each nodes selection.  This results in reading in the wrong
data from disk for distributed datasets.  The fix uses the shape of the
dataset pre-selection.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,9 @@ jobs:
       run: |
         mpirun -np 4 pytest caput/tests/test_memh5_parallel.py
         mpirun -np 4 pytest caput/tests/test_mpiarray.py
+        mpirun -np 1 pytest caput/tests/test_selection_parallel.py
+        mpirun -np 2 pytest caput/tests/test_selection_parallel.py
+        mpirun -np 4 pytest caput/tests/test_selection_parallel.py
 
   build-docs:
 
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1059,13 +1059,20 @@ def _len_slice(slice_, n):
 
 def _reslice(slice_, n, subslice):
     # For a slice along an axis of length n, return the slice that would select the
-    # start:end elements of the final array
+    # slice(start, end) elements of the final array.
+    #
+    # In other words find a single slice that has the same affect as application of two successive
+    # slices
     dstart, dstop, dstep = slice_.indices(n)
 
     if subslice.step is not None and subslice.step > 1:
         raise ValueError("stride > 1 not supported. subslice: %s" % subslice)
 
-    return slice(dstart + subslice.start * dstep, dstart + subslice.stop * dstep, dstep)
+    return slice(
+        dstart + subslice.start * dstep,
+        min(dstart + subslice.stop * dstep, dstop),
+        dstep,
+    )
 
 
 def _expand_sel(sel, naxis):
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -656,7 +656,7 @@ class MPIArray(np.ndarray):
             # Loop over partitions of the IO and perform them
             for part in partitions:
                 islice, fslice = _partition_sel(
-                    sel, split_axis, gshape[split_axis], part
+                    sel, split_axis, dshape[split_axis], part
                 )
                 dist_arr[fslice] = dset[islice]
 
--- a/caput/tests/test_selection.py
+++ b/caput/tests/test_selection.py
@@ -1,3 +1,10 @@
+"""Serial version of the selection tests."""
+# === Start Python 2/3 compatibility
+from __future__ import absolute_import, division, print_function, unicode_literals
+from future.builtins import *  # noqa  pylint: disable=W0401, W0614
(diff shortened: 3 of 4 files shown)


ssiegelx:
This fixes a bug where the wrong data would be read from disk when performing a selection on a distributed dataset.  I noticed this when attempting to down select along the frequency axis when reading in some of the pipeline products.  The freq axis in the index map contains the requested frequencies, but the actual data is from completely different frequencies.  The following code snippet reproduces the error:

```
import os
import numpy as np
from draco.core import containers

freq = np.array([x for x in zip(np.linspace(800.0, 400.0, 1024, endpoint=False), np.repeat(0.390625, 1024))],
                dtype=[('centre', np.float32), ('width', np.float32)])

pix = np.arange(1)

nfreq = freq.size
npix = pix.size

freq_sel = slice(500, 600)

# Create a container with a dataset that just contains the frequency index
data = containers.Map(freq=freq, pixel=pix)
data.map[:] = np.arange(nfreq)[:, np.newaxis, np.newaxis]

# Write to disk
filename = os.path.expanduser("~/freq_sel_test.h5")

data.save(filename)

# Read from disk (not distributed)
test = containers.Map.from_file(filename, distributed=False, freq_sel=freq_sel)
print(np.any(test.map[:] != data.map[freq_sel]))
print(test.map[:, 0, 0])
```

```
False
[500. 501. 502. 503. 504. 505. 506. 507. 508. 509. 510. 511. 512. 513.
 514. 515. 516. 517. 518. 519. 520. 521. 522. 523. 524. 525. 526. 527.
 528. 529. 530. 531. 532. 533. 534. 535. 536. 537. 538. 539. 540. 541.
 542. 543. 544. 545. 546. 547. 548. 549. 550. 551. 552. 553. 554. 555.
 556. 557. 558. 559. 560. 561. 562. 563. 564. 565. 566. 567. 568. 569.
 570. 571. 572. 573. 574. 575. 576. 577. 578. 579. 580. 581. 582. 583.
 584. 585. 586. 587. 588. 589. 590. 591. 592. 593. 594. 595. 596. 597.
 598. 599.]
```

```
# Read from disk (distributed)
test_dist = containers.Map.from_file(filename, distributed=True, freq_sel=freq_sel)
print(np.any(test_dist.map[:] != data.map[freq_sel]))
print(test_dist.map[:, 0, 0])
```

```
True
[100. 101. 102. 103. 104. 105. 106. 107. 108. 109. 110. 111. 112. 113.
 114. 115. 116. 117. 118. 119. 120. 121. 122. 123. 124. 125. 126. 127.
 128. 129. 130. 131. 132. 133. 134. 135. 136. 137. 138. 139. 140. 141.
 142. 143. 144. 145. 146. 147. 148. 149. 150. 151. 152. 153. 154. 155.
 156. 157. 158. 159. 160. 161. 162. 163. 164. 165. 166. 167. 168. 169.
 170. 171. 172. 173. 174. 175. 176. 177. 178. 179. 180. 181. 182. 183.
 184. 185. 186. 187. 188. 189. 190. 191. 192. 193. 194. 195. 196. 197.
 198. 199.]
```

This affects selections made on ch_pipeline and draco containers, and also frequency selections made when using the from_acq_h5_fast method of andata.CorrData.  andata.CorrData uses its own code for doing a slow read along the frequency axis, so it's not effected by this bug.

jrs65:
Yikes. This is a pretty bad bug! Thanks for catching and fixing it.

It's bad that our tests didn't catch it. @nritsche is away, so I can try and quickly write a test that does catch it, and then we can get this merged. I was about to start using this functionality.

jrs65:
@ssiegelx could you take a quick look over the new test to check if it looks good to you? It did fail without your patch, and it passes with it, so that's something.

ssiegelx:
I guess the original parallel test passed because the datasets did not have the `__memh5_distributed_dset` attribute, so it just used the serial version of the code? 

This looks good, but the test still passes for me pre-patch when run with 1 process (I have not tried running it with more than 1 process).  I think it only fails for certain types of selections.  For example, it fails pre-patch when `fsel = slice(5, 8, 1)`.  The relevant logic regarding what will fail and what will pass is in `mpiarray._reslice`.

jrs65:
> I guess the original parallel test passed because the datasets did not have the `__memh5_distributed_dset` attribute, so it just used the serial version of the code?

Yup, that was one part, and also that they didn't even run multiple processes under MPI so it wasn't doing a very thorough test.
 
> This looks good, but the test still passes for me pre-patch when run with 1 process (I have not tried running it with more than 1 process). I think it only fails for certain types of selections. For example, it fails pre-patch when `fsel = slice(5, 8, 1)`. The relevant logic regarding what will fail and what will pass is in `mpiarray._reslice`.

Good catch. I think there's a bug in `_reslice` where it doesn't actually respect the upper bound of the initial slice. That does not cause any problems when the code is running correctly, but it has made this bug harder to understand. I'll also try and fix that one now.

jrs65:
I've expanded the tests to run on different numbers of MPI processes (1, 2, 4) and to try reading a variety of different selections. That should cover all the cases now. I'll merge when the tests pass.

PR #140, opened 2020-09-04
Docs build

nritsche:


jrs65:
Is the documentation viewable somewhere? I only get the old version on RTD.

Also, why remove the custom autodoc templates? I think they were needed to produce reasonable listings.

jrs65:
Also, it's probably playing around with what we use to render the actual docstrings. I think we are currently using `numpydoc` but `napoleon` which does a similar thing was upstreamed into Sphinx. Although I think I played with this before and I have this feeling it might produce something which doesn't quite work.

nritsche:
> Is the documentation viewable somewhere? I only get the old version on RTD.
> 
> Also, why remove the custom autodoc templates? I think they were needed to produce reasonable listings.

https://caput.readthedocs.io/en/rn-rtd/

jrs65:
Thanks.

Okay, here's a few comments. 

- The left sidebar always seems to show "Configuration" for me, no matter which page I'm on.
- I can't click through to method documentation, e.g. click on `memh5`, and click on `MemDatasetDistributed`, I should be able to click on `redistribute` to get access to the extended method docs (which does exist), but there's no link. I think that's what the `autosummary` templates are for, i.e. to generate the pages for those entries.

nritsche:
>     * The left sidebar always seems to show "Configuration" for me, no matter which page I'm on.

Because that's the only top level page. I'd agree that the start/overview page also appears there.

>     * I can't click through to method documentation, e.g. click on `memh5`, and click on `MemDatasetDistributed`, I should be able to click on `redistribute` to get access to the extended method docs (which does exist), but there's no link. I think that's what the `autosummary` templates are for, i.e. to generate the pages for those entries.

Maybe that's what it once did. If I build with those templates, it produces this now:
Screenshot from 2020-12-21 12-52-36

jrs65:
Here's another oddity on this page: https://caput.readthedocs.io/en/rn-rtd/generated/caput.time.html#module-caput.time

For some reason I dumps the full rendered docstrings for some of the entries in a toctree table.

image

jrs65 (approved):
I added a comment about some weird issues. If you can't figure out what's going on in 10 mins, I'd say we should just merge and open an issue to fix it. Having anything new on RTD is clearly better than what we currently have.

nritsche:
- [ ] weighted median docs are not getting built

nritsche:
https://caput--140.org.readthedocs.build/en/140/index.html

PR #142, opened 2020-09-30
perf(weighted_median): use quickselect to calculate the static median

jrs65:
This moves to using a weighted quickselect algorithm for calculating the
*static* weighted median which is significantly faster (> 50x) than the
original AVL tree code.

### DIFF
--- a/caput/weighted_median.pyx
+++ b/caput/weighted_median.pyx
@@ -9,6 +9,347 @@ from MedianTree cimport Tree, Data
 cimport numpy as np
 cimport cython
 
+# Define the fused types that can be used for the data or weights in the median routine
+ctypedef fused data_t:
+    int
+    long
+    float
+    double
+
+ctypedef fused weight_t:
+    int
+    long
+    float
+    double
+
+
+@cython.wraparound(False)
+@cython.boundscheck(False)
+cdef data_t _quickselect_weight(data_t[::1] A, weight_t[::1] W, double qs, char method):
+    # A, W: data and weights respectively, both are modified in place
+    # qs: the amount of weight that we are selecting the element at
+    # method: one of l, h or s to indicate whether to take the lower, higher or split
+    #         (average) value if there is a tie
+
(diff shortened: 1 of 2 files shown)


jrs65:
@nritsche any hint with the readthedocs error?

nritsche:
Oh, I must have activated docs builds for this, too. But it won't work. We can ignore it though, it's not used for branch protection yet.

I started reviewing (and github made a mess out of the diff). What was in your force push?

jrs65:
Nothing really. I had enabled a pytest plugin that will run doctests for Cython code, but it doesn't work with the newest Pytest and so I ended up disabling it for now.

caput/weighted_median.pyx
```diff
@@ -9,99 +9,366 @@ from MedianTree cimport Tree, Data
...
+        raise TypeError("Type of data array is not supported.")
 
-    if data.ndim is 1:
-        return _weighted_median_1D(data, weights, c_method)
-    if data.ndim is 2:
-        return _weighted_median_2D(data, weights, c_method)
-    raise NotImplementedError('weighted_median is only implemented for 1 and 2 dimensions, not {}'
-                              .format(data.ndim))
+    if A.ndim == 1:
+        return res[0]
```
nritsche, line 313: I don't understand what this case is about.

nritsche (approved):
This looks pretty perfect.

My review is mostly questions and comments I added after I understood what something does.

caput/weighted_median.pyx
jrs65: It's so that the routine returns a scalar if the input array was only 1D.

nritsche: Oh yeah, it does make sense :+1:

jrs65:
> Oh, I must have activated docs builds for this, too. But it won't work. We can ignore it though, it's not used for branch protection yet.

I've deactivated the readthedocs check for the time being, so hopefully we'll get rid of those red X's

PR #143, opened 2020-09-30
Support zero length array arguments for caput.time routines

jrs65:


jrs65:
Apparently this doesn't work on the newest skyfield. I've filed a bug report: skyfielders/python-skyfield#458

nritsche:
> Apparently this doesn't work on the newest skyfield. I've filed a bug report: [skyfielders/python-skyfield#458](https://github.com/skyfielders/python-skyfield/issues/458)

Should we set a maximum version for the skyfield requirement for now?

jrs65:
That's probably a good idea. I don't really like setting max versions in general because at some point you end up needing to upgrade and have to deal with a large amount of changes all at once, better to get them every now and again, however, skyfield is a bit ridiculous at the moment.

I'm going to defer on this for a few days, as it's possible there'll be a fix (and I could also workaround it fairly easily), but partly because if we pin a version for a while I'd like to figure out exactly which version, and so I want to look at the other public API changes that broke the code you're using that I want to put into caput.

caput/time.py
```diff
@@ -672,21 +678,23 @@ def ensure_unix(time):
...
-            if isinstance(time0, timelib.Time):
+            if isinstance(time[0], timelib.Time):
                 return datetime_to_unix(time.utc_datetime())
         except ImportError:
             pass
 
         # Finally try and convert into a float.
         try:
-            return np.float64(time)
+            if np.issubdtype(time.dtype, np.number):
```
nritsche, line 688: Do you not want to return anything if this is false?

nritsche (approved):

caput/time.py
nritsche: Also, the try-catch can probably be smaller.

jrs65:
@nritsche I changed the skyfield call I did to something using the more public API. Can you take another look?

caput/tests/test_time.py
```diff
@@ -12,348 +11,360 @@
...
-    mirror_url = "https://bao.chimenet.ca/skyfield/"
+files = ["Leap_Second.dat", "deltat.data", "deltat.preds", "de421.bsp"]
+
+for file in files:
+    ctime.skyfield_wrapper.load(mirror_url + file)
+
+
+def test_epoch():
+
+    from skyfield import earthlib
```
nritsche, line 38: This was already imported.

caput/tests/test_time.py
```diff
@@ -12,348 +11,360 @@
...
+
+    # Tolerance limited by stellar aberation
+    assert lst == approx(TRA, abs=0.01, rel=1e-10)
+
+
+def test_transit_array():
+    # Do a simple test of transit_RA in an array. Use the fact that the RA
+    # advances predictably to predict the answers
+
+    from skyfield import earthlib
```
nritsche, line 63: Already imported

caput/tests/test_time.py
```diff
@@ -12,348 +11,360 @@
...
-        unix_epoch = ctime.datetime_to_unix(epoch)
-        TRA = obs.transit_RA(unix_epoch)
+def test_lsa_skyfield():
+    # Check an lsa calculated by caput.time against one calculated by PyEphem
 
-        # Calculate LST
-        t = ctime.unix_to_skyfield_time(unix_epoch)
-        gst = earthlib.sidereal_time(t)
-        lst = (360.0 * gst / 24.0 + obs.longitude) % 360.0
+    from skyfield import earthlib, api
```
nritsche, line 112: `earthlib` is already imported

nritsche (approved):
:+1: 
I just found out you can click on "force pushed" and see the changes.

jrs65:
> 👍
> I just found out you can click on "force pushed" and see the changes.

That is amazing. It's going to revolutionize PR reviews!

PR #144, opened 2020-10-14
Pipeline profiling

jrs65:


tristpinsm (approved):
Sweet! This is going to be really convenient to use. And `pyinstrument` seems like a nice utility. It was much quicker to review than I had expected.

It makes sense to use this in an interactive session, and I imagine in the vast majority of cases this will be sufficient, but if we ever wanted to profile a larger scale job `queue` could be made to parse these parameters from the config. Might not be worth bothering with that until it comes up though.

jrs65:
Agreed. It wouldn't be hard to pass this one through the yaml files, but I looked at doing it, and figured it was going to take me a little too long for it to be worth it. As you suggest, it shouldn't be hard to do the next time it seems to be needed.

In the mean time if you do want to profile a batch job, it's pretty easy to run `caput-pipeline queue --no-submit <file.yaml>`, then find the constructed `jobscript.sh` file and edit the line to make it `caput-pipeline run --profile ...`

PR #146, opened 2020-11-10
Skyfield refactor and transit/rise/set time routines

jrs65:
This is the start of a refactoring of all the ephemeris code.

### DIFF
--- a/caput/tests/test_time.py
+++ b/caput/tests/test_time.py
@@ -217,6 +217,23 @@ def test_lsd_array():
     # Check that it works with zero length arrays
     assert obs.lsd_to_unix(np.array([])).size == 0
 
+    # Check that is works with lists (this was previously a bug)
+    assert obs.lsd_zero() == approx(obs.lsd_to_unix([0.0, 0.0]), abs=1e-3, rel=0)
+
+
+def test_era_accuracy():
+
+    # Pick a time to check the ERA around
+    dts = ctime.ensure_unix(datetime(2000, 1, 1))
+
+    # These should give back the same time, but the accuracy of the STELLAR_S constant
+    # and Skyfields interpolation of dUT1 limit this.
+    t0 = ctime.era_to_unix(0, dts)
+    t1 = ctime.era_to_unix(0, dts - 5 * 3600)
+
+    # The accuracy should be better than a millisecond
+    assert t0 == approx(t1, abs=1e-3)
+
 
 def test_datetime_to_string():
     dt = datetime(2014, 4, 21, 16, 33, 12, 12356)
@@ -362,3 +379,5 @@ def test_ensure_unix():
 
     # Check that it works for zero length arrays
     assert ctime.ensure_unix(np.array([])).size == 0
+
+
--- a/caput/tests/test_time.py
+++ b/caput/tests/test_time.py
@@ -48,14 +48,14 @@ def test_epoch():
     # Calculate the transit_RA
     unix_epoch = ctime.datetime_to_unix(epoch)
     TRA = obs.transit_RA(unix_epoch)
+    LSA = obs.lsa(unix_epoch)
 
     # Calculate LST
-    t = ctime.unix_to_skyfield_time(unix_epoch)
-    gst = earthlib.sidereal_time(t)
-    lst = (360.0 * gst / 24.0 + obs.longitude) % 360.0
+    lst = obs.lst(unix_epoch)
 
     # Tolerance limited by stellar aberation
     assert lst == approx(TRA, abs=0.01, rel=1e-10)
+    assert lst == approx(LSA, abs=0.01, rel=1e-10)
 
 
 def test_transit_array():
@@ -68,9 +68,7 @@ def test_transit_array():
     obs = ctime.Observer(118.3, 36.1)
 
     # Calculate LST
-    t = ctime.unix_to_skyfield_time(ctime.datetime_to_unix(epoch))
-    gst = earthlib.sidereal_time(t)
-    lst = (360.0 * gst / 24.0 + obs.longitude) % 360.0
+    lst = obs.lst(ctime.datetime_to_unix(epoch))
 
     # Drift rate should be very close to 1 degree/4minutes.
     # Fetch times calculated by ephem
@@ -79,11 +77,13 @@ def test_transit_array():
 
     # Calculate RA using transit_RA
     unix_epoch = ctime.datetime_to_unix(epoch)
-    unix_times = unix_epoch + (delta_deg * 60 * 4 * ctime.SIDEREAL_S)
+    unix_times = unix_epoch + (delta_deg * 60 * 4 * ctime.STELLAR_S)
     TRA = obs.transit_RA(unix_times)
+    LSA = obs.lsa(unix_times)
 
     # Compare
     assert lst == approx(TRA, abs=0.02, rel=1e-10)
+    assert lst == approx(LSA, abs=0.02, rel=1e-10)
 
 
 def test_delta():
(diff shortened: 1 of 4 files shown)


caput/time.py
```diff
@@ -408,6 +412,220 @@ def transit_RA(self, time):
...
+        Returns
+        -------
+        times : np.ndarray
+            UNIX times of transits.
+        dec : np.ndarray
+            Only returned if `return_dec` is set. Declination of source at transit.
+        """
+
+        # The function to find routes for. For the upper transit we just search for
+        # HA=0, for the lower transit we need to rotate the 180 -> -180 transition to
```
nritsche, line 449: What's HA?

caput/time.py
```diff
@@ -408,6 +412,220 @@ def transit_RA(self, time):
...
+        def f(t):
+            ha = self._source_ha(source, t)
+            return ha if not lower else (ha % 360.0) - 180.0
+
+        # Calculate the HA at the initial search points
+
+        # Get the ends of the search interval in TT Julian days
+        t0 = ensure_unix(t0)
+        t1 = ensure_unix(t1) if t1 is not None else t0 + 24 * 3600.0
+        t0jd, t1jd = unix_to_skyfield_time([t0, t1]).tt
```
nritsche, line 460: images

nritsche (commented):
I instantly know it's your code, when I see a variable name like `t0jd`. :smile: 
Mostly I just attached some small comments. :+1: 

Should there be a `breaking changes` section to dca8f12 ?

Was there something in bondia that can make use of this change when it's in? I remember we redefined something there provisionally.

I think it's good @fandinomat also has a look.

caput/time.py
```diff
@@ -408,6 +412,220 @@ def transit_RA(self, time):
...
         return ra
 
+    def transit_times(
+        self, source, t0, t1=None, step=0.2, lower=False, return_dec=False
+    ):
+        """Find the transit times of the given source in an interval.
+
+        Parameters
+        ----------
+        source : skyfield source
```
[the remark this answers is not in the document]
jrs65: The problem is it refers to a whole bunch of classes that expose the same interface. `skyfield source` is about the best overarching description.

caput/time.py
jrs65: Hour angle. That one should be very well understood by astronomers.

caput/time.py
```diff
@@ -387,28 +404,232 @@ def transit_RA(self, time):
...
+            UNIX times of transits.
+        dec : np.ndarray
+            Only returned if `return_dec` is set. Declination of source at transit.
+        """
+
+        # The function to find routes for. For the upper transit we just search for
+        # HA=0, for the lower transit we need to rotate the 180 -> -180 transition to
+        # be at 0.
+        def f(t):
+            ha = self._source_ha(source, t)
```
nritsche, line 452:
```suggestion
            # hour angle
            ha = self._source_ha(source, t)
```

caput/time.py
```diff
@@ -387,28 +404,232 @@ def transit_RA(self, time):
...
 
+    def transit_times(
+        self, source, t0, t1=None, step=0.2, lower=False, return_dec=False
+    ):
+        """Find the transit times of the given source in an interval.
+
+        Parameters
+        ----------
+        source : skyfield source
+            The source we are calculating the transit of.
```
nritsche, line 423:
```suggestion
            The source we are calculating the transit of. This could for example be a .. or ...
```

caput/tests/test_time.py
```diff
@@ -362,3 +383,55 @@ def test_ensure_unix():
...
+
+    times = chime.transit_times(eph["sun"], dts, dte)
+
+    # Calculated via the old version of `ch_util.ephemeris.solar_transit(dts, dte)``
+    precalc_times = [1604605326.0967, 1604691728.9071]
+
+    # Check that the calculations agree within 2s. This criterion comes from the first
+    # attempts to use the new routines, and seems reasonable enough that I'm not going
+    # to track down the difference
+    assert times == approx(precalc_times, abs=2)
```
fandinomat, line 413: Don't you need np.all() here?

caput/tests/test_time.py
```diff
@@ -362,3 +383,55 @@ def test_ensure_unix():
...
+    times, risings = chime.rise_set_times(eph["sun"], dts, dte)
+
+    assert times == approx(precalc_times, abs=2)
+    assert np.all(risings == precalc_risings)
+
+    risings = chime.rise_times(eph["sun"], dts, dte)
+    settings = chime.set_times(eph["sun"], dts, dte)
+
+    assert risings == approx(precalc_times[precalc_risings], abs=2)
+    assert settings == approx(precalc_times[~precalc_risings], abs=2)
```
fandinomat, line 437: Similarly, don't you need np.all() in all four assert statements above?

caput/time.py
```diff
@@ -863,16 +1081,71 @@ def ephemeris(self):
...
+
+    roots = []
+    increasing = []
+
+    # Search through intervals
+    for xa, xb, fa, fb in zip(x_init[:-1], x_init[1:], f_init[:-1], f_init[1:]):
+
+        # Entries are the same sign, so there is no solution in between.
+        # NOTE: we need to deal with the case where one edge might be an exact root,
+        # hence the strictly greater than 0.0
```
fandinomat, line 1132: Does `brentq` accept one edge being the root? Or did you mean to deal with this case sepparately and forgot?

fandinomat (approved):
Just a few comments/questions. But I guess if the tests are passing, then you probably don't need the np.all() calls I suggested.

caput/time.py
jrs65: I'm glad you can recognize my "style". I don't think this one is that bad, it just indicates that t0jd is the same time as t0 just expressed in Julian Days (for which the standard abbreviation is JD)! I'll add a comment to make things clearer.

caput/tests/test_time.py
jrs65: Actually pytest takes care of that for you when you use `pytest.approx`.

caput/tests/test_time.py
jrs65: See above.

caput/time.py
jrs65: It does indeed. You only need to take care of it when you're trying to detect regions which contain a solution.

caput/tests/test_time.py
```diff
@@ -217,6 +220,24 @@ def test_lsd_array():
     # Check that it works with zero length arrays
     assert obs.lsd_to_unix(np.array([])).size == 0
 
+    # Check that is works with lists (this was previously a bug)
+    assert obs.lsd_zero() == approx(obs.lsd_to_unix([0.0, 0.0]), abs=1e-3, rel=0)
+    obs.lsd_to_unix
```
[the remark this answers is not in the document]
jrs65: Good catch.

caput/time.py
```diff
@@ -995,12 +994,18 @@ class SkyfieldWrapper(object):
...
 
-    def __init__(self, path=None, expire=False, ephemeris="de421.bsp"):
+    def __init__(self, path=None, expire=None, ephemeris="de421.bsp"):
 
         import os
 
         self._ephemeris_name = ephemeris
 
+        if expire is not None:
+            warnings.warn(
```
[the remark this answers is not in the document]
nritsche: > Maybe I should throw an exception instead if it's been explicitly set?
 Yeah, I guess that makes sense then.

caput/time.py
```diff
@@ -1098,14 +1081,8 @@ def ephemeris(self):
...
 
-        exp_val = self.load.expire
-        self.load.expire = True
-
-        try:
-            self.timescale
-            self.ephemeris
-        finally:
-            self.load.expire = exp_val
+        self.load.download("finals2000A.all")
```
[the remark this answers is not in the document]
nritsche: Oups. yeah, not the same file

PR #150, opened 2020-12-21
fix(transit_times): Handle small intervals better

### LINKED ISSUE #149: time.transit_times produces incorrect results with non-sensical interval and step size combinations
Passing `time.transit_times` a small enough time range without also adjusting the step size (unsurprisingly) results in no transit being found:
```
>>> import datetime
>>> from caput.time import Observer
>>> Observer(lon=-119.62, lat=49.32, alt=545.0,lsd_start=datetime.datetime(2013, 11, 15)).transit_times(
        299.86815, 1.60660594e+09, 1.60660596e+09)
array([], dtype=float64)
```

Adjusting the step size to a reasonable value fixes this:
```
>>> Observer(lon=-119.62, lat=49.32, alt=545.0,lsd_start=datetime.datetime(2013, 11, 15)).transit_times(
        299.86815, 1.60660594e+09, 1.60660596e+09, step=4.6296296296296294e-05) 
array([1.60660595e+09])
```

This is the cause of a regression in `ch_util` (https://github.com/chime-experiment/ch_util/issues/40)

This function should raise an error if `step` is larger than the interval.  And I'd argue the best thing to do would be to make the default step size smaller (say, a fifth of the interval) in cases when the supplied interval is less than a day.

ketiltrout:
If `step` is not provided, `transit_times` will use a default initial search step of one fifth of the search interval, or 0.2 days, whichever is smaller.

Also:
 - raises `ValueError` if the initial search step is larger than the search interval, in which case no transit will ever be found.
 - raises `ValueError` if the search interval is non-positive. Without this, `NumPy` instead raises `ValueError` with a cryptic message complaining about negative array length)

Closes #149

### DIFF
--- a/caput/tests/test_time.py
+++ b/caput/tests/test_time.py
@@ -14,7 +14,7 @@ from datetime import datetime
 import numpy as np
 import pytest
 from skyfield import earthlib, api
-from pytest import approx
+from pytest import approx, raises
 
 from caput import time as ctime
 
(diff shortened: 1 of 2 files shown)


jrs65 (approved):
These look great. However, I think the rise/set routines will also suffer from the same bug. Could you possibly fix them too? I think all you need to do is copy your fix into `Observer._sr_work`. Thanks!

ketiltrout:
I've done you one better and eliminated the duplicate code

jrs65 (approved):
Awesome. Thanks Don!

PR #151, opened 2020-12-21
Remove Python 2 support

jrs65:


jrs65:
@nritsche @fandinomat I've tagged you for review. I'll squash all the commits together when you're done, but for the moment, I figured it might make it easier to review.

caput/misc.py
```diff
@@ -9,20 +9,16 @@
...
 
 import collections
 import importlib
-import inspect
 import os
 
+# TODO: remove this. This was to support a patching of this routine to support Python 2
+# that used to exist in here. This will be removed when all other repos are changed to
+# use the version from `inspect`
+from inspect import getfullargspec
```
nritsche, line 20: Let's explicitly warn about this being deprecated when it's used:
for example
```
from deprecated import deprecated

@property
@deprecated(version='21.1.0', reason="""
    Use inspect.getfullargspec instead.
    This was to support a patching of this routine to support Python 2 that 
    used to exist in here and will be removed in version 21.6.0
""")
def getfullargspec()
    return inspect.getfullargspec
```

Or you could also use https://docs.python.org/3/library/exceptions.html#DeprecationWarning

nritsche (approved):

ketiltrout (approved):

fandinomat (approved):

caput/misc.py
jrs65: Done

PR #154, opened 2021-01-13
fix(time.Observer.transit_times): Use sidereal day.

tristpinsm:
Use the sidereal second to set the end time to one sidereal day
later if not specified. Otherwise you get two transit times if
t0 is within a few minutes of transit.

This seemed like unexpected behaviour to me.

### DIFF
--- a/caput/time.py
+++ b/caput/time.py
@@ -584,7 +584,7 @@ class Observer(object):
         # Get the ends of the search interval
         t0 = ensure_unix(t0)
         if t1 is None:
-            t1 = t0 + 24 * 3600.0
+            t1 = t0 + 24 * 3600.0 * STELLAR_S
         else:
             t1 = ensure_unix(t1)
             if t1 <= t0:


caput/time.py
```diff
@@ -566,7 +566,7 @@ def _fixup_interval_and_step(self, t0, t1, step):
         # Get the ends of the search interval
         t0 = ensure_unix(t0)
         if t1 is None:
-            t1 = t0 + 24 * 3600.0
+            t1 = t0 + 24 * 3600.0 * SIDEREAL_S
```
jrs65, line 569: This probably should be STELLAR_S

jrs65 (changes requested):

caput/time.py
jrs65: Otherwise, I think you make a fair point!

tristpinsm: back to wikipedia I go...

jrs65: Do you agree it should be stellar day in the end?

tristpinsm: I think so. If we've specified RA and Dec in ICRS coordinates we should use that reference to calculate the next transit. I could test it by comparing the predicted time between the next two transits to the stellar/diereal day. Practically, it's only an 8ms difference.

jrs65 (approved):

tristpinsm:
I compared the predicted difference between transit times of CygA ("predicted length of day") by `caput.time.Observer.transit_times` to the sidereal and stellar days in this plot: 
image

In conclusion I think stellar day makes sense to use here. I'm not sure what is responsible for the oscillations in the predicted times. Maybe nutation? the NRAO website says there is a nutation mode on half a year period, but I don't know the magnitude of that effect.

PR #157, opened 2021-01-14
CI: Run flake8 and pylint

nritsche:
- Activate flake8&pylint checks without enforcing it
- Fix a few of the warnings


jrs65:
I'm not sure how many of the style things we should turn off for pylint. It would be nice not to make things worse, but there's quite a lot of errors getting thrown. It seems like ones we could turn off would be:
- missing-function-docstring
- protected-access
- no-member (seems to generate a lot of false positives)
- fixme

But for the first three it would definitely be better if we didn't make them worse.

jrs65 (approved):

nritsche:
> I'm not sure how many of the style things we should turn off for pylint. It would be nice not to make things worse, but there's quite a lot of errors getting thrown. It seems like ones we could turn off would be:
> 
>     * missing-function-docstring
> 
>     * protected-access
> 
>     * no-member (seems to generate a lot of false positives)
> 
>     * fixme
> 
> 
> But for the first three it would definitely be better if we didn't make them worse.

The only option (other than fixing everything) I see for not-making-them-worse is adding`# pylint: ignore W*` comments. And I don't think we want that...

jrs65:
Actually @nritsche this looks to be absolutely perfect for us: https://pypi.org/project/pylint-ignore/

nritsche:
> Actually @nritsche this looks to be absolutely perfect for us: https://pypi.org/project/pylint-ignore/

Thanks. I made use of this and I fixed a few more warnings.

nritsche:
Oups. This needs more work: I will add submodules to caput.config

jrs65 (approved):

PR #160, opened 2021-01-19
Fix linter

### LINKED ISSUE #159: Linting issues
Two related issues:

First, running the `caput-pipeline lint` command breaks with a `venv` config key. It seems to be confused about whether it's dealing with a string or `Path` object
```
caput-pipeline lint qs_singledays.yaml
Activating '/home/jrs65/chime_pipeline/venv/'...
Traceback (most recent call last):
  File "/home/jrs65/chime_pipeline/venv/bin/caput-pipeline", line 11, in <module>
    load_entry_point('caput', 'console_scripts', 'caput-pipeline')()
  File "/home/jrs65/chime_pipeline/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jrs65/chime_pipeline/venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/jrs65/chime_pipeline/venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jrs65/chime_pipeline/venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jrs65/chime_pipeline/venv/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/jrs65/chime_pipeline/code/caput/caput/scripts/runner.py", line 44, in lint_config
    load_venv(f)
  File "/home/jrs65/chime_pipeline/code/caput/caput/scripts/runner.py", line 75, in load_venv
    if not base.exists():
AttributeError: 'str' object has no attribute 'exists'
```

Second, the queue command despite "linting" the file before submitting doesn't break with this because it doesn't run the linter within in the venv.

nritsche:
Closes #159

### DIFF
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -1,16 +1,9 @@
-from os.path import (
-    normpath,
-    expanduser,
-    expandvars,
-    realpath,
-    abspath,
-    isabs,
-    join,
-    exists,
-)
+import os
+import sys
+
+from pathlib import Path
 
 import click
-import sys
 
 from caput.config import CaputConfigError
 
@@ -69,17 +62,12 @@ def load_venv(configfile):
 
     click.echo("Activating '{}'...".format(venv_path))
 
-    # TODO: python2 - use pathlib
-    import os.path
-
-    base = os.path.abspath(venv_path)
+    base = Path(venv_path).resolve()
     if not base.exists():
         click.echo("Path defined in 'cluster'/'venv' doesn't exist ({})".format(base))
         sys.exit(1)
 
-    site_packages = os.path.join(
-        base, "lib", "python{}".format(sys.version[:3]), "site-packages"
-    )
+    site_packages = base / "lib" / "python{}".format(sys.version[:3]) / "site-packages"
     prev_sys_path = list(sys.path)
 
     site.addsitedir(site_packages)
@@ -234,7 +222,6 @@ def queue(configfile, submit=False, lint=True):
         same filesystem.
     """
 
-    import os.path
     import shutil
     import yaml
 
@@ -279,12 +266,12 @@ def queue(configfile, submit=False, lint=True):
     # If the system is specified update the current config with it
     if "system" in conf:
 
-        sys = conf["system"]
+        system = conf["system"]
 
-        if sys not in system_defaults:
-            raise ValueError('Specified system "%s": is not known.' % sys)
+        if system not in system_defaults:
+            raise ValueError('Specified system "%s": is not known.' % system)
 
-        rconf.update(**system_defaults[sys])
+        rconf.update(**system_defaults[system])
 
     # Update the current config with the rest of the users variables
     rconf.update(**conf)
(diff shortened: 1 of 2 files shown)


caput/scripts/runner.py
```diff
@@ -68,15 +62,12 @@ def load_venv(configfile):
...
-    base = os.path.abspath(venv_path)
+    base = Path(venv_path).resolve()
     if not base.exists():
         click.echo("Path defined in 'cluster'/'venv' doesn't exist ({})".format(base))
         sys.exit(1)
 
-    site_packages = os.path.join(
+    site_packages = PurePath.joinpath(
         base, "lib", "python{}".format(sys.version[:3]), "site-packages"
     )
```
jrs65, line 72:
```suggestion
    site_packages = base / "lib" / f"python{sys.version[:3])} / "site-packages"
```

caput/scripts/runner.py
```diff
@@ -314,9 +323,9 @@ def queue(configfile, submit=False, lint=True):
     rconf["workdir"] = workdir
     rconf["finaldir"] = finaldir
     rconf["scriptpath"] = fixpath(__file__)
-    rconf["logpath"] = join(jobdir, "jobout.log")
-    rconf["configpath"] = join(jobdir, "config.yaml")
-    rconf["statuspath"] = join(jobdir, "STATUS")
+    rconf["logpath"] = PurePath.joinpath(jobdir, "jobout.log")
+    rconf["configpath"] = PurePath.joinpath(jobdir, "config.yaml")
+    rconf["statuspath"] = PurePath.joinpath(jobdir, "STATUS")
```
jrs65, line 328: Presumably when the formatting is done below, these Path objects get converted into strings when the `script % rconf` is evaulated.
```suggestion
    rconf["logpath"] = jobdir / "jobout.log"
    rconf["configpath"] = jobdir / "config.yaml"
    rconf["statuspath"] = jobdir / "STATUS"
```

caput/scripts/runner.py
```diff
@@ -281,29 +290,29 @@ def queue(configfile, submit=False, lint=True):
...
     finaldir = expandpath(rconf["directory"])
-    if not isabs(finaldir):
+    if not PurePath.is_absolute(finaldir):
         raise ValueError("Final output directory path %s must be absolute" % finaldir)
 
     # Create temporary directory if required
-    jobdir = join(workdir, "job/")
-    if not exists(jobdir):
-        os.makedirs(jobdir)
+    jobdir = PurePath.joinpath(workdir, "job/")
```
jrs65, line 302:
```suggestion
    jobdir = workdir / "job/"
```

caput/scripts/runner.py
```diff
@@ -281,29 +290,29 @@ def queue(configfile, submit=False, lint=True):
...
     # Construct the working directory
     workdir = expandpath(rconf["temp_directory"])
-    if not isabs(workdir):
+    if not PurePath.is_absolute(workdir):
         raise ValueError("Working directory path %s must be absolute" % workdir)
 
     # Construct the output directory
     finaldir = expandpath(rconf["directory"])
-    if not isabs(finaldir):
+    if not PurePath.is_absolute(finaldir):
```
jrs65, line 298:
```suggestion
    if not finaldir.is_absolute():
```

caput/scripts/runner.py
```diff
@@ -1,16 +1,11 @@
...
-    realpath,
-    abspath,
-    isabs,
-    join,
-    exists,
-)
+import os
+import sys
+
+from pathlib import Path, PurePath
```
jrs65, line 4: You probably don't need to be importing `PurePath` directly here. Removing the import will help identify places you should be calling methods on the objects directly.

caput/scripts/runner.py
```diff
@@ -34,7 +29,6 @@ def cli():
 )
 def lint_config(configfile):
```
jrs65, line 30: I was surprised to learn that `click.Path` arguments don't return a `pathlib.Path` object. You'd think they would, but they can't because they need to support Python 2 apparently.

jrs65 (changes requested):
Thanks!

This all looks like it should work fine to me now. However, I think the use of pathlib in here could do with straightening out. There's a bunch of places where pathlib methods are being called like functions, e.g. `Path.is_absolute(f)` rather than `f.is_absolute()`, and where `Path.joinpath(a, b)` is being used instead of `a / b`. I've made a few suggestions, but they're not exhaustive so you'll probably need to look through more carefully.

caput/scripts/runner.py
nritsche: Yes, I tried that, they get converted as expected.

jrs65 (approved):

PR #161, opened 2021-01-21
fix(scripts.runner): Broken pathlib usage.

tristpinsm:
Fixes some problems introduced by previous pathlib commit:
- expandpath would raise an exception if path doesn't exist,
  breaking with previous behaviour.
- mkdir call for job directory didn't create parent path
- use of + with PosixPath

I didn't exhaustively audit that commit. I only fixed things until I could get `chp item generate` to work again.


jrs65:
Thanks Tristan!

jrs65 (approved):

PR #162, opened 2021-02-01
feat(mpiarray): ufunc, __array_finalize__, and  __getitem__ handling for MPIArray

anjakefala:
If used with draco, requires: https://github.com/radiocosmology/draco/pull/125

# Subclassing NumPy arrays 101
`view(MPIArray)` -> `__array_finalize__`
`MPIArray[slice]` -> `__getitem__` -> `__array_finalize__`
`MPIArray()` -> `__new__` -> `__array_finalize__`

ufuncs are the universal functions that are applied element by element in nparrays. Such a `np.add()`, `np.multiply()`. If they end up summing over an axis, they are a ufunc with a `reduce` method. If they go element-by-element, they are a ufunc with an `outer` method. If they occur in-place, they have an `at` method.

ufunc -> `__array_ufunc__`

More links on the role all these various functions play in writing subclasses for NumPy arrays can be found here: https://github.com/chime-experiment/Pipeline/issues/81

## New Exceptions

The new `AxisException` will be added. It will be raised when there are issues involving the integrity of the distributed axis with MPIArrays.

## __getitem__
- When you use `global_slice` to index into a distributed axis, then it will return an array for the rank on which the index exists, and `None` otherwise.
- When you slice on a non-distributed axis, `__getitem__` will return an MPIArray, whose distributed axes number might be lower than the original array, depending on if the slice results in an axis reduction. It is assumed that the distributed axis length is unchanged. An `mpiutil.split_local` call will be made.
- When you directly slice into a distributed axis, it will index into the local arrays on each rank, and return a regular numpy array. This behaviour will be deprecated, and replaced with the raising of an `AxisException`.

## ufunc

- Upon a ufunc being executed, both inputs and outputs are converted to
standard nparrays, and then the nparray ufunc is called
- all inputted MPIArrays *must* be distributed along the same axes. An `AxisException` will be called, if they are not.
- if no output are provided, the results are converted back to MPIArrays. The new array will either be distributed over that axis, or possibly one axis down for `reduce` methods. `keepdims` kwargs are handled.
- If you pass a kwarg `axis` to the ufunc, it must not be the distributed axis. Operations along the distributed axis are not allowed.
- For operations that normally return a scalar, the scalars will be wrapped into a 1D array, distributed across axis 0.
- To produce the MPIArray output, `MPIArray.wrap(..)` is called. This means there is a call to `mpiutil.split_local` and `mpiutil.allreduce`.

## __array_finalize__

This is called whenever a user uses `.view()`, `__new__`, and broadcasts on an MPIArray. It finalizes the creation of the output MPIArray. `view()`s can occur with `__getitem__` calls.

If it is a `__new__` call, it does nothing.

If we are in an `np.ndarray.view()` call, it does nothing. This should only occur when we are within a `wrap()`

If we are in a `view()`, it grabs the attributes from the origin array. 

## Misc

If a user wishes to create an MPIArray from an ndarray, they should use `MPIArray.wrap()`. They should not use `ndarray.view(MPIArray)`.

https://github.com/chime-experiment/Pipeline/issues/81

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -55,7 +55,8 @@ Its behaviour changes depending on the exact slice it gets:
 It's important to note that it never communicates data between ranks. It only
 ever operates on data held on the current rank.
 
-Example
+Global Slicing Examples
+-----------------------
 
 Here is an example of this in action::
 
@@ -98,7 +99,8 @@ Direct Slicing
 This can be used for both fetching and assignment. It is recommended
 to only index into the non-parallel axis or to do a full slice `[:]`.
 
-Behaviour
+Direct Slicing Behaviour
+------------------------
 
 - A full slice `[:]` will return a :class:`MPIArray` on fetching, with
   identical properties to the original array.
@@ -108,7 +110,8 @@ Behaviour
 - Any indexing into the parallel axis will result into a local index on each rank,
   returning a regular `numpy` array.
 
-Example
+Direct Slicing Examples
+-----------------------
 
     import numpy as np
     from caput import mpiarray, mpiutil
@@ -137,12 +140,14 @@ In NumPy, universal functions (or ufuncs) are functions that operate on ndarrays
 in an element-by-element fashion. :class:`MPIArray` supports all ufunc calculations,
 except along the parallel axis.
 
-Requirements
+ufunc Requirements
+------------------
 
 - All input :class:`MPIArray` *must* be distributed along the same axis.
 - If you pass a kwarg `axis` to the ufunc, it must not be the parallel axis.
 
-Behaviour
+ufunc Behaviour
+---------------
 
 - If no output are provided, the results are converted back to MPIArrays. The new array
   will either be parallel over the same axis as the input MPIArrays, or possibly one axis
@@ -152,7 +157,8 @@ Behaviour
   parallel across axis 0.
 - shape related attributes will be re-calculated
 
-Example
+ufunc Examples
+--------------
 
     import numpy as np
     from caput import mpiarray, mpiutil
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -253,13 +253,20 @@ class MPIArray(np.ndarray):
         if (dist_axis_index != slice(None, None, None)) and (
             dist_axis_index != slice(0, self.local_array.shape[self.axis], None)
         ):
+            import warnings
             if isinstance(dist_axis_index, int):
-                raise AxisException(
-                    "Cannot directly index distributed axis; use global_slice instead"
+                warnings.warn(
+                    "You are indexing directly into the distributed axis."
+                    "Returning a view into the local array."
+                    "Please use global_slice, or .local_array before indexing instead."
                 )
-            raise AxisException(
-                "Cannot directly sub-slice distributed axis; use global_slice instead"
-            )
+
+                return self.local_array.__getitem__(v)
+            warnings.warn(
+                    "You are directly sub-slicing the distributed axis."
+                    "Returning a view into the local array."
+                    "Please use global_slice, or .local_array before indexing.")
+            return self.local_array.__getitem__(v)
 
         # Figure out which is the axis number for the distributed axis after the slicing
         # by removing slice axes which are just ints from the mapping.
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -254,6 +254,7 @@ class MPIArray(np.ndarray):
             dist_axis_index != slice(0, self.local_array.shape[self.axis], None)
         ):
             import warnings
+
             if isinstance(dist_axis_index, int):
                 warnings.warn(
                     "You are indexing directly into the distributed axis."
@@ -263,9 +264,10 @@ class MPIArray(np.ndarray):
 
                 return self.local_array.__getitem__(v)
             warnings.warn(
-                    "You are directly sub-slicing the distributed axis."
-                    "Returning a view into the local array."
-                    "Please use global_slice, or .local_array before indexing.")
+                "You are directly sub-slicing the distributed axis."
+                "Returning a view into the local array."
+                "Please use global_slice, or .local_array before indexing."
+            )
             return self.local_array.__getitem__(v)
 
         # Figure out which is the axis number for the distributed axis after the slicing
@@ -1253,7 +1255,7 @@ class MPIArray(np.ndarray):
 
         # that ufunc was not implemented for ndarrays
         if results is NotImplemented:
-            raise NotImplemented
+            raise NotImplementedError
 
         # operation was performed in-place, so we can just return
         if method == "at":
@@ -1266,10 +1268,8 @@ class MPIArray(np.ndarray):
             # reduction methods eliminate axes, so the distributed axis
             # might need to be recalculated
             # except when the user explicitly specifies keepdims
-            if not kwargs.get("keepdims", False):
-                reduction_axis = kwargs["axis"]
-                if reduction_axis < dist_axis:
-                    dist_axis -= 1
+            if not kwargs.get("keepdims", False) and (kwargs["axis"] < dist_axis):
+                dist_axis -= 1
 
         ret = []
 
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -214,15 +214,17 @@ class _global_resolver:
             slobj = tuple(slice(None, None, None) if sl is None else sl for sl in slobj)
 
             # Return an MPIArray view
-            arr = self.array[slobj]
-
             # Figure out which is the distributed axis after the slicing, by
             # removing slice axes which are just ints from the mapping
             dist_axis = [
                 index for index, sl in enumerate(slobj) if not isinstance(sl, int)
             ].index(self.axis)
 
-            return MPIArray.wrap(arr, axis=dist_axis, comm=self.array._comm)
+            return MPIArray.wrap(
+                self.array.view(np.ndarray)[slobj],
+                axis=dist_axis,
+                comm=self.array._comm,
+            )
 
     def __setitem__(self, slobj, value):
 
@@ -1194,7 +1196,6 @@ class MPIArray(np.ndarray):
                 if reduction_axis < dist_axis:
                     dist_axis -= 1
 
-        # Wrapping the results back into MPIArrays, distributed across the appropriate axis
         ret = []
 
         for result, output in zip(results, outputs):
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1093,10 +1093,10 @@ class MPIArray(np.ndarray):
         for array in inputs:
             if isinstance(array, MPIArray):
                 if dist_axis is None:
-                    dist_axis = array._axis
+                    dist_axis = array.axis
                 else:
                     assert (
-                        dist_axis == array._axis
+                        dist_axis == array.axis
                     ), "The distributed axis for all MPIArrays in an expression should be the same"
 
                 args.append(array.local_array.view(np.ndarray))
@@ -1151,7 +1151,14 @@ class MPIArray(np.ndarray):
         else:
             outputs = (None,) * ufunc.nout
 
-        results = super().__array_ufunc__(ufunc, method, *args, **kwargs)
+        results = super().__array_ufunc__(ufunc, method, *args, **kwargs) # pylint: disable=no-member
+
+        if results is NotImplemented:
+            return NotImplemented
+
+        # operation is performed in-place, so we can just return
+        if method == 'at':
+            return
 
         if ufunc.nout == 1:
             results = (results,)
@@ -1179,7 +1186,7 @@ class MPIArray(np.ndarray):
                         MPIArray.wrap(np.reshape(result, (1, 1)), axis=0)
                     )
 
-        return ret[0] if ufunc.nout == 1 else tuple(ret)
+        return ret[0] if len(ret) == 1 else tuple(ret)
 
     def __array_finalize__(self, obj):
         """
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1264,7 +1264,9 @@ class MPIArray(np.ndarray):
         if ufunc.nout == 1:
             results = (results,)
 
-        if "reduce" in method and results[0].shape:
+        if "reduce" in method and (
+            results[0].shape or getattr(outputs[0], "shape", None)
+        ):
             # reduction methods eliminate axes, so the distributed axis
             # might need to be recalculated
             # except when the user explicitly specifies keepdims
@@ -1276,6 +1278,10 @@ class MPIArray(np.ndarray):
         for result, output in zip(results, outputs):
             # case: results were placed in the array specified by `out`; return as is
             if output is not None:
+                if hasattr(output, "axis") and output.axis != dist_axis:
+                    raise AxisException(
+                        f"provided output MPIArray's distributed axis is not consistent with expected output distributed axis. Expected {dist_axis}; Actual {output.axis}"
+                    )
                 ret.append(output)
             else:
                 # case: the result is an ndarray; wrap it into an MPIArray
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1226,16 +1226,14 @@ class MPIArray(np.ndarray):
         if obj is None:
             return
 
-        # we are in a ufunc, rebuild the attributes from the original MPIArray
-        if isinstance(obj, MPIArray):
-            comm = getattr(obj, "comm", mpiutil.world)
-
-        if hasattr(obj, "axis"):
-            axis = obj.axis
-        else:
-            # in the middle of an np.ndarray.view()
-            # user will have to set the attributes themselves
+        if not isinstance(obj, MPIArray):
+            # in the middle of an np.ndarray.view() in the wrap()
             return
+
+        # we are in a slice, rebuild the attributes from the original MPIArray
+        comm = getattr(obj, "comm", mpiutil.world)
+
+        axis = obj.axis
 
         # Get shape and offset
         lshape = self.shape
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1233,7 +1233,9 @@ class MPIArray(np.ndarray):
         if hasattr(obj, "axis"):
             axis = obj.axis
         else:
-            raise Exception("Cannot construct output axis")
+            # in the middle of an np.ndarray.view()
+            # user will have to set the attributes themselves
+            return
 
         # Get shape and offset
         lshape = self.shape
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -844,6 +844,40 @@ class MPIArray(np.ndarray):
 
         return dist_arr
 
+    def allreduce(self, op=None):
+        """Perform MPI reduction operation `op` by pickling local array.
+
+        Return results for all ranks.
+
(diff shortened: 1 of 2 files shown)


jrs65:
@anjakefala we touched on this yesterday, but what I think would help a lot would be to collect a list of MPIArray calls that we would like to work and agree on what their behaviour should be (for which we might want to circulate around to chime-analysis for opinions). Then we can turn them all into unit test cases.

Something like:
```python
# Setup array
comm = MPI.COMM_WORLD
dist_array = mpiarray.MPIArray((comm.size, 4), comm=comm, axis=0)
dist_array[:] = comm.rank

# I think this one is uncontroversial
assert dist_array.sum(axis=1) == 4 * comm.rank

# Should this be allowed?
# assert dist_array.sum(axis=0) == ???

# What should this do? It seems the two sensible options are that is sums over all axes, giving the same scalar everywhere...
assert dist_array.sum() == 4 * comm.size * (comm.size - 1) // 2

# ... or that it just ignores the distributed axis, giving a distributed 1-d array
assert dist_array.sum() == 4 * comm.rank
```

caput/tests/test_mpiarray.py
```diff
@@ -394,7 +397,69 @@ def test_global_setslice(self):
...
+
+        dist_arr_scalar = mpiarray.MPIArray((size, 4), axis=0)
+        dist_arr_scalar[:] = rank * 2
+
+        dist_arr_add = dist_arr + dist_arr
+
+        # Check that you can add two numpy arrays,
+        # if they are distributed along the same axes
+        # Check that you can multiple a numpy array against a scalar
+        assert (dist_arr_add == dist_arr_scalar).all()
```
jrs65, line 415: This test doesn't check the scalar case afaict. The `rank * 2` above is just a multiplication of two python scalars which are then assigned into the MPIArray.

I would also recommend separating out these two tests into separate asserts, i.e.
```suggestion
        dist_arr_add = dist_arr + dist_arr
        disr_arr_mul = 2 * dist_arr

        # Check that you can add two MPIArrays
        assert (dist_arr_add == 2 * rank).all()
        
        # Check that you can multiple an MPIArray against a scalar
        assert (dist_arr_sum == 2 * rank).all()
```

It's probably worth checking it works with differing arguments too (i.e. add different MPIArrays).

Also I think you were working on this, but we should test the exceptions in cases you expect to fail (i.e. different distributed axes, different shapes...),

jrs65, line 415: Also, note that `.all()` in an MPIArray reduction over all axes, and we hadn't clearly worked out the semantics for that yet.

Seeing how it works with `.all()` maybe pushes it in the direction of you should get one value per rank.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+        In NumPy, ufuncs are the various fundamental operations applied to ndarrays in an element-by-element fashion, such as add() and divide().
+        https://numpy.org/doc/stable/reference/ufuncs.html
+
+        ndarray has lots of built-in ufuncs. In order to use them, the MPIArrays need to be converted into ndarrays, otherwise NumPy reports a NotImplemented error.
+
+        The distributed axis for all input MPIArrays, is expected to be the same. Operations across the distributed axis, will not be permitted.
+
+        The new array will ehter be be distributed over that axis, or possibly one axis down for `reduce` methods, if they eliminate the distributed axis.
+
+        For operations that normally return a scalar, the scalars will be wrapped into a 1D array, distributed across axis 0.
```
jrs65, line 1091: Please wrap lines in docstrings at the same length as the source.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+        dist_axis = None
+
+        for array in inputs:
+            if isinstance(array, MPIArray):
+                if dist_axis is None:
+                    dist_axis = array.axis
+                else:
+                    assert (
+                        dist_axis == array.axis
+                    ), "The distributed axis for all MPIArrays in an expression should be the same"
```
jrs65, line 1067: Probably avoid using an `assert` which can be bypassed by run time options. Should explicitly raise a ValueError.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+        # will be set when we return to it
+        if obj is None:
+            return
+
+        # we are in a ufunc, rebuild the attributes from the original MPIArray
+        if isinstance(obj, MPIArray):
+            comm = getattr(obj, "comm", mpiutil.world)
+            axis = getattr(
+                obj, "axis", 0
+            )  # probably not a good default! How would we find this out?
```
jrs65, line 1187: If you can't reconstruct the distributed axis then it should throw an exception rather than guess.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+
+            # if the array has been reduced, re-calibrate the distr axes
+            if len(self.shape) < axis + 1:
+                axis = len(self.shape) - 1
```
jrs65, line 1191: I don't think this clause is correct, as it doesn't seem to depend on which axis was the reduction axis.

I would check your test coverage here. I think something like this should fail with the clause as above.
```
dist_arr = MPIArray((3, 3, 3, 3), comm=MPI.COMM_WORLD, axis=1)
dist_arr.sum(axis=0)
```
I think the issue is that the arrays in the tests below aren't long enough to trigger this error.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+        if method == "at":
+            return
+
+        if ufunc.nout == 1:
+            results = (results,)
+
+        if "reduce" in method:
+            # reduction methods eliminate axes, so the distributed axes needs to be recalculated
+            if results[0].shape and len(results[0].shape) < dist_axis + 1:
+                dist_axis = len(results[0].shape) - 1
```
jrs65, line 1142: See the comment below about this clause.

caput/mpiarray.py
```diff
@@ -1036,6 +1036,177 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+
+        # Wrapping the results back into MPIArrays, distributed across the appropriate axis
+        ret = []
+
+        for result, output in zip(results, outputs):
+            if (
+                output is not None
+            ):  # case: results were placed in the array specified by `out`; return as is
```
jrs65, line 1150: This and the clause below are where black seems to have made the code less readable. I would suggest just putting the comment within the block below the clause.

jrs65 (changes requested):

anjakefala:
Linter is failing due to pylint updates. This PR is ready for another conversation now.

caput/mpiarray.py
```diff
@@ -214,15 +214,23 @@ def __getitem__(self, slobj):
...
+            if dist_axis != self.axis:
+
+                return MPIArray.wrap(
+                    self.array.view(np.ndarray)[slobj],
+                    axis=dist_axis,
+                    comm=self.array._comm,
+                )
+            else:
+                return MPIArray.wrap(
+                    self.array[slobj], axis=dist_axis, comm=self.array._comm
```
anjakefala, line 232: It is better to reduce the number of hidden MPI calls, and so it would be better to do this within an MPIArray `__getitem__`, and remove `wraps`. If the issue is the calculation of the `totallength` of the distributed axes, we can prevent operations which adjust the size of the total length, and grab it from the global shape of the original axes..

anjakefala:
Ready for the next review.

caput/mpiarray.py
```diff
@@ -245,6 +236,71 @@ class MPIArray(np.ndarray):
         The dimension to distribute the array across.
     """
 
+    def __getitem__(self, v):
+        # Return an MPIArray view
+
+        if not isinstance(v, tuple):
+            return super().__getitem__(v)
```
jrs65, line 243: What happens for this?
```
dist_arr = mpiarray.MPIArray((4,), axis=0, comm=comm)
a = dist_arr[0]
```
I think this should fail because it indexes the distributed axis, but because of this if statement it would actually work because it just calls the numpy implementation.

caput/mpiarray.py
```diff
@@ -245,6 +236,71 @@ class MPIArray(np.ndarray):
...
+
+        try:
+            dist_axis_index = v[self.axis]
```
jrs65, line 246: I don't think this works. What if I run this through it?
```
dist_arr = mpiarray.MPIArray((4,4), axis=1, comm=comm)
a = dist_arr[0]
```
I think it'll fail with an IndexError.

jrs65 (changes requested):
This is looking great. Things are looking very neat.

I just had a quick skim through, so here's a few suggested changes. I think it would help to have some more thorough testing to catch issues like these, as there's probably a few more issues still to find.

caput/mpiarray.py
```diff
@@ -1069,6 +1153,203 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+            raise AxisException(
+                f"operations along the distributed axis (in this case, {dist_axis}) are not allowed."
+            )
+
+        # 'out' kwargs contain arrays that the ufunc places the results into
+        # this views the local part of the output arrays into an ndarray
+        # that the ufunc knows how to work with
+        outputs = kwargs.get("out", None)
+        if outputs:
+            out_args, _ = MPIArray._mpi_to_ndarray(outputs)
```
jrs65, line 1249: Do you need to check consistency of the distributed axis between the outputs and the inputs?

caput/mpiarray.py
```diff
@@ -1069,6 +1153,203 @@ def _partition_io(self, skip=False, threshold=1.99):
...
+            kwargs["out"] = tuple(out_args)
+        else:
+            outputs = (None,) * ufunc.nout
+
+        results = super().__array_ufunc__(ufunc, method, *args, **kwargs)
+
+        # that ufunc was not implemented for ndarrays
+        if results is NotImplemented:
+            raise NotImplementedError
```
jrs65, line 1258: I'm not sure this is correct, although the numpy docs aren't super clear. I think either you want to raise a concrete instance of the exception `NotImplementedError` (i.e. `raise NotImplementedError("...")`), or `return NotImplemented`.

caput/tests/test_mpiarray.py
```diff
@@ -289,14 +293,18 @@ def test_global_getslice(self):
         res = local_array[:, 3:5]
 
         assert isinstance(arr, mpiarray.MPIArray)
+        assert arr.axis == 0
```
jrs65, line 296: This file seems to be an awkward mix of pytest (e.g. bare asserts) and unittest (e.g. self.assertRaisesWhenTheMoonIsFull) style test cases. I think the former supports the latter but not the other way around, so I'd probably recommend converting the entire file to pytest style.

There don't appear to be any setup/teardown methods so it should be very quick to port as you don't need to write any pytest fixtures.

caput/mpiarray.py
```diff
@@ -1069,6 +1153,203 @@ def _partition_io(self, skip=False, threshold=1.99):
         slices = [slice(start, end) for start, end in zip(starts, ends)]
         return split_axis, slices
 
+    @staticmethod
```
jrs65, line 1156: This should probably just be a module level function. I think generally accepted style in Python is to avoid staticmethod's unless you have a good reason.

caput/mpiarray.py
```diff
@@ -245,6 +237,92 @@ class MPIArray(np.ndarray):
         The dimension to distribute the array across.
     """
 
+    def __getitem__(self, v):
```
jrs65, line 240: Can you add a bit of documentation about what is and isn't supported in MPIArray? in terms of ufuncs and slicing etc, I think it would help people to have a clear description of that.

jrs65: Probably in the class level docstring.

jrs65: It doesn't need crazy detail. But something would be good. A few examples wouldn't hurt too, although you could just point people to the tests.

jrs65 (changes requested):

caput/mpiarray.py
anjakefala: In practice, it is hard to do from the inputs alone (the axis might change over the course of the operation), but I can check if the expected dist axis matches the dist axis of the outputs!

caput/tests/test_mpiarray.py
anjakefala: I was using the `unittest.assertRaises` to be able to test that an exception occurred when it was supposed to. I want to check if pytest has something that does something similar.

anjakefala: Examples:

```
         # check that subtracting arrays with two different distributed axis fails                                                                                                                              
         self.assertRaises(                                                                                                                                                                                     
             mpiarray.AxisException,                                                                                                                                                                            
             np.subtract,                                                                                                                                                                                       
             mpiarray.MPIArray((size, 4), axis=0),                                                                                                                                                              
             mpiarray.MPIArray((size, 4), axis=1),                                                                                                                                                              
         )                                                                                                                                                                                                      
                                                                                                                                                                                                                
         # check that outer ufunc on arrays that cannot be broadcast fails                                                                                                                                      
         self.assertRaises(                                                                                                                                                                                     
             ValueError,                                                                                                                                                                                        
             np.multiply,                                                                                                                                                                                       
             mpiarray.MPIArray((size, 3), axis=0),                                                                                                                                                              
             mpiarray.MPIArray((size, 4), axis=0),                                                                                                                                                              
         )                                                                                                                                                                                                      
```

anjakefala: It seems it does support it: https://docs.pytest.org/en/stable/assert.html

jrs65: Indeed it does. I think you mostly need to change that and then change all the class methods into functions.

caput/mpiarray.py
anjakefala: Yes! I put these in some of the method docstrings, but I will write them in the class.

caput/mpiarray.py
```diff
@@ -89,6 +92,100 @@
             print(ri, arr3)
         mpiutil.barrier()
 
+Direct Slicing
+==============
+
+The :class: `MPIArray` supports direct slicing, with `__getitem__`.
```
jrs65, line 98:
```suggestion
The class `MPIArray` supports direct slicing, with `MPIArray.__getitem__`.
```
I'm sort of guessing at your intent here, but i) I don't think sphinx will like the space between the class and the name, and ii) you usually don't need the type specifiers for things within the same file.

jrs65: Also I think this line might confuse some of our less Python-savvy team members into thinking they need to call `__getitem__` directly. Maybe just say it supports slicing using `[...]` as usual and that it's implemented by `__getitem__`.

caput/mpiarray.py
```diff
@@ -34,6 +34,9 @@
     # Perform the lag transform on the frequency direction.
     darr4 = MPIArray.wrap(np.fft.irfft(darr3, axis=0), axis=1)
 
+Note: If a user wishes to create an MPIArray from an ndarray, they should use MPIArray.wrap(). They should not use ndarray.view(MPIArray).
```
jrs65, line 37: Ironically this line about `.wrap` needs to be wrapped.

caput/mpiarray.py
```diff
@@ -89,6 +92,100 @@
...
+- Any indexing or slicing into the non-parallel axis, will return a
+:class: `MPIArray`. The number associated with the parallel axis,
+will be adjusted if a slice results in an axis reduction.
+- Any indexing into the parallel axis will result into a local index on each rank,
+returning a regular `numpy` array.
+
+Example
+-------
+
+    import numpy as np
```
jrs65, line 115: I would suggest that you make this example runnable via doctest (see `caput.config` for an example). That way we'll get some bonus testing, but more importantly it means that the example will work and won't end up out of date. Practically that means putting `>>>` in front of some lines, putting a `True` under the test lines and you'll need to make it agnostic on the communicator size (as doctest will only run with a size=1), but that should be as easy as just replacing the 4 with `mpiutil.size`.

caput/mpiarray.py
```diff
@@ -89,6 +92,100 @@
...
+- If no output are provided, the results are converted back to MPIArrays. The new array
+will either be parallel over the same axis as the input MPIArrays, or possibly one axis
+down if the `ufunc` is applied via a `reduce` method (i.e. the shape of the array is reduced
+by one axis).
+- For operations that normally return a scalar, the scalars will be wrapped into a 1D array,
+parallel across axis 0.
+- shape related attributes will be re-calculated
+
+Example
+------
```
jrs65, line 156:
```suggestion
-------
```

caput/mpiarray.py
```diff
@@ -89,6 +92,100 @@
...
+
+    import numpy as np
+    from caput import mpiarray, mpiutil
+
+    dist_arr = mpiarray.MPIArray((mpituil.size, 4), axis=0)
```
jrs65, line 161: A practical example of why doctest is useful...
```suggestion
    dist_arr = mpiarray.MPIArray((mpiutil.size, 4), axis=0)
```

caput/mpiarray.py
jrs65, line 156: Also doctest me!

jrs65 (changes requested):
@anjakefala this is really great, and the new documentation is excellent. I have a few small suggestions, but the major one is to make the examples runnable via doctest (and make sure they are run in the CI)

jrs65 (approved):
I think we're done. Awesome work! I don't want to merge quite yet as I just want to think about what else we need to do to ensure it won't immediately break a bunch of existing code in the pipeline.

jrs65:
Obviously you'll also want to rebase this down to a saner set of commits!

caput/tests/test_mpiarray.py
```diff
@@ -583,14 +581,11 @@ def test_reduce():
...
+        # an ndarray vector with appropriate values + shape
+        assert (dist_array.allreduce() == 4).all()
+        assert (dist_array.allreduce()).shape == (4, 1)
 
-        with pytest.raises(ValueError):
-            # ValueError should be raised, since df_total and dist_array are not the correct shape
-            dist_array.Allreduce(df_total)
+        # MPIArray.sum().allreduce() should give the scalar sum of
+        # all entries
+        assert df_sum.allreduce() == 4 * size
```
jrs65, line 591: I don't think this will work if the number of elements on each rank is different, i.e.
```
assert comm.size == 4
dist_arr = MPIArray((5,), comm=comm)

dist_arr.allreduce()
```
So I would suggest adding a test for it and seeing if it will break.

jrs65 (changes requested):

caput/tests/test_mpiarray.py
anjakefala: That case does not work with the regular `Allreduce` either!

You get: `mpi4py.MPI.Exception: MPI_ERR_TRUNCATE: message truncated` with both situations.

How would you like the code to handle this? Perform a check, and print a more readable error message?

anjakefala: From a zoom call:

the case we are most interested in is `array.sum().allreduce()`, so we are deciding to restrict usage of `allreduce()` to arrays with a single element.

jrs65 (approved):

anjakefala:
:sob:

PR #163, opened 2021-02-04
fix(OneAndAone): move validation to special function

### LINKED ISSUE #148: Bug in caput.pipeline
When running the attached pipeline files you get this exception. This happens because the `__init__` for `OneAndOne` does some validation, but it requires the keys and queues to be setup, which are now not created by the manager until after task initialisation.

@cubranic created this minimal example for the issue.

```
Traceback (most recent call last):
  File "/usr/local/Caskroom/miniconda/base/envs/caput/bin/caput-pipeline", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/scripts/runner.py", line 81, in run
    P = Manager.from_yaml_file(configfile)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 517, in from_yaml_file
    return cls.from_yaml_str(yaml_doc)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 552, in from_yaml_str
    self._setup_tasks()
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 624, in _setup_tasks
    task, key_spec = self._setup_task(task_spec)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 701, in _setup_task
    task = task_cls._from_config(task_params)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 852, in _from_config
    self.__init__()
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 1303, in __init__
    _OneAndOne.__init__(self)
  File "/usr/local/Caskroom/miniconda/base/envs/caput/lib/python3.9/site-packages/caput/pipeline.py", line 1096, in __init__
    if len(self._in) != n_args:
AttributeError: 'CookEggs' object has no attribute '_in'
```

Classes in `new_eggs.py`:
```python
from caput.pipeline import *
from caput import config


class PrintEggs(TaskBase):

    eggs = config.Property(proptype=list)

    def __init__(self):
        self.i = 0

    def setup(self):
        print("Setting up PrintEggs.")

    def next(self):
        if self.i >= len(self.eggs):
            raise PipelineStopIteration()
        print("Spam and %s eggs." % self.eggs[self.i])
        self.i += 1

    def finish(self):
        print("Finished PrintEggs.")


class GetEggs(TaskBase):

    eggs = config.Property(proptype=list)

    def __init__(self):
        self.i = 0

    def setup(self):
        print("Setting up GetEggs.")

    def next(self):
        if self.i >= len(self.eggs):
            raise PipelineStopIteration()
        egg = self.eggs[self.i]
        self.i += 1
        return egg

    def finish(self):
        print("Finished GetEggs.")


class CookEggs(IterBase):

    style = config.Property(proptype=str)

    def setup(self):
        print("Setting up CookEggs.")

    # def next(self, egg):
    #     print("Cooking %s %s eggs." % (self.style, egg))

    def process(self, egg):
        print("Cooking %s %s eggs." % (self.style, egg))

    def finish(self):
        print("Finished CookEggs.")
```

run with this pipeline config:
```yaml
---
pipeline:
  tasks:
    - type: new_eggs.PrintEggs
      params: eggs_params

    - type: new_eggs.GetEggs
      params: eggs_params
      out: egg

    - type: new_eggs.CookEggs
      params: cook_params
      in: egg

eggs_params:
  eggs: ['green', 'duck', 'ostrich']

cook_params:
  style: 'fried'
```

nritsche:
~Set up in, out and requires keys in constructor instead of in Manager.add_task.~
Do validation of special keys in `validate()`, instead of in constructor

Closes #148

### DIFF
--- a/caput/mpiutil.py
+++ b/caput/mpiutil.py
@@ -313,7 +313,7 @@ def split_m(n, m):
     base = n // m
     rem = n % m
 
-    part = base * np.ones(m, dtype=np.int) + (np.arange(m) < rem).astype(np.int)
+    part = base * np.ones(m, dtype=int) + (np.arange(m) < rem).astype(int)
 
     bound = np.cumsum(np.insert(part, 0, 0))
 
--- a/caput/pipeline.py
+++ b/caput/pipeline.py
@@ -63,6 +63,7 @@ trivial but fully implemented task:
 ...     eggs = config.Property(proptype=list)
 ...
 ...     def __init__(self):
+...         super().__init__()
 ...         self.i = 0
 ...
 ...     def setup(self):
@@ -88,6 +89,7 @@ that are designed to operate in this manner.
 ...     eggs = config.Property(proptype=list)
 ...
 ...     def __init__(self):
+...         super().__init__()
 ...         self.i = 0
 ...
 ...     def setup(self):
@@ -780,6 +782,9 @@ class Manager(config.Reader):
             )
             raise config.CaputConfigError(msg) from e
 
+        # The tasks own custom validation method
+        task.validate()
+
         self.tasks.append(task)
         logger.debug(f"Added {task.__class__.__name__} to task list.")
 
@@ -829,6 +834,9 @@ class TaskBase(config.Reader):
         """
 
         pass
+
+    def validate(self):
+        """Validate the task after instantiation."""
 
     def next(self, input=None):
         """Iterative analysis stage of pipeline task.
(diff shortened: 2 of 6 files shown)


jrs65:
@nritsche it seems like this breaks the ability to construct tasks by hand in Python (see the docs on pure Python pipelines), or at least it makes it a lot harder as you need to pass the keys yourself to the init.

jrs65:
At a broader level I think this starts to blur what was a fairly clean separation of concerns in the API, i.e. the tasks care only about the single thing they do, and the Manager knows how to connect things together and pass them around.

At a more practical level, I fear that this is just going to break a lot of code that assumes that it is the top level `__init__`. The code shouldn't do that, but in practice who puts a super init call in you know that you're the top level init (until you're not).

Was there some reason that the existing framework couldn't be minimally changed to fix this bug?

nritsche:
> @nritsche it seems like this breaks the ability to construct tasks by hand in Python (see the docs on pure Python pipelines), or at least it makes it a lot harder as you need to pass the keys yourself to the init.

Maybe it can go to _from_config instead then?

jrs65:
It's a difficult one. In my opinion the problem is with the `__init__` of `_OneAndOne` which makes an assumption that it is being constructed by the Manager, and thus has all the special properties set. Ultimately that whole task seems problematic, as it tries to hide a lot of serialising and deserialising inputs/outputs to disk behind the scenes, and in my opinion that's a mistake (but it's a bit of a biased opinion).

It only needs the special attributes that the Manager inserts to do some validation, so one way out I can imagine is that we could have `TaskBase._validate()` method that subclasses are allowed to implement if they want to do any special validation before they run. In that routine they are allowed to assume they've been added by the Manager and thus have all the special attributes. That `.validate()` method would probably be called on the task in `Manager.add_task(...)`

jrs65 (approved):

PR #164, opened 2021-02-12
Profiling improvements for caput.pipeline

nritsche:
- [x] Depends on https://github.com/radiocosmology/caput/pull/163

### DIFF
--- a/caput/misc/__init__.py
+++ b/caput/misc/__init__.py
@@ -1,15 +1,9 @@
 """A set of miscellaneous routines that don't really fit anywhere more specific."""
 
-import collections
 import importlib
-import logging
 import os
-import time
 
 import numpy as np
-import psutil
-
-logger = logging.getLogger(__name__)
 
 
 def vectorize(**base_kwargs):
@@ -265,7 +259,7 @@ class lock_file:
     Examples
     --------
 
-    >>> from . import memh5
+    >>> from .. import memh5
     >>> container = memh5.BasicCont()
     >>> with lock_file('file_to_create.h5') as fname:
     ...     container.save(fname)
@@ -332,7 +326,7 @@ def getfullargspec(*args, **kwargs):
     import warnings
 
     warnings.warn(
-        "This patch to support Python 2 is no longer needed and will be removed.",
+        "This patch to support Python 2 is no longer needed and will be removed after December 2021.",
         DeprecationWarning,
     )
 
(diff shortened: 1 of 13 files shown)


caput/misc/__init__.py
```diff
@@ -46,15 +45,15 @@ def __call__(self, *args, **kwargs):
 
             return arr
 
-        def __get__(self, obj, type=None):
+        def __get__(self, obj, type_=None):
```
anjakefala, line 48: TIL that the norm is to append an `_` (not prepend, which is what I thought it was.

nritsche: prepend on members means "please treat this as private", so it would confuse to use that

jrs65: Hmmm... I don't typically mind examples like this that mask builtin names. I mean, we're all adults and we can see that it's fine in here. However, as a compromise how about we just rename it to something else without an underscore. The python docs on descriptors refer to this parameter as `objtype` which seems good enough to me, and doesn't have the awkward trailing underscore.

caput/misc/profiling.py
```diff
@@ -0,0 +1,137 @@
+"""Profiling for the caput pipeline"""
```
jrs65, line 1: How about just puttomg this under a `caput.profiling` module, rather than making `caput.misc` a subpackage?

caput/misc/profiling.py
```diff
@@ -0,0 +1,137 @@
...
+            Description of what is profiled. You have to pass the same str to `stop`.
+        """
+        self._start_time[name] = time.time()
+
+        # Get all stats at the same time
+        with self.oneshot():
+            self._start_cpu_times[name] = self.cpu_times()
+            self.cpu_percent()
+            self._start_memory[name] = self.memory_full_info().uss
+            self._start_disk_io[name] = self.io_counters()
```
jrs65, line 56: This will break on macOS at `Process.io_counters` doesn't exist.

caput/scripts/runner.py
```diff
@@ -317,6 +334,10 @@ def queue(configfile, submit=False, lint=True):
     else:
         rconf["venv"] = "/dev/null"
 
+    # Forward profiler configuration
+    rconf["profile"] = "--profile" if profile else ""
+    rconf["profiler"] = f"--profiler={profiler}"
```
jrs65, line 339:
```suggestion
    rconf["profiler"] = f"--profiler={profiler}" if profile else ""
```

caput/scripts/runner.py
```diff
@@ -135,8 +137,9 @@ def run(configfile, loglevel, profile, profiler):
             pr = Profiler()
             pr.start()
 
+    use_psutil_profiler = profile & (profiler == "psutil")
```
jrs65, line 140:
```suggestion
    use_psutil_profiler = profile and (profiler == "psutil")
```

caput/pipeline.py
```diff
@@ -967,6 +1015,23 @@ def _setup_keys(self, in_=None, out=None, requires=None):
...
+        """
+        Get state of the task.
+
+        Returns
+        -------
+        pipeline_state : str
+            One of 'setup', 'next', 'finish' or 'raise'.
+        """
+        return self._pipeline_state
+
```
jrs65, line 1034: Why bother? These are only used internally within the pipeline framework. I don't think we want them to be exposed to a public level.

caput/misc/profiling.py
```diff
@@ -0,0 +1,137 @@
...
+import logging
+import time
+
+import numpy as np
+import psutil
+
+logger = logging.getLogger(__name__)
+
+
+class PSUtilProfiler(psutil.Process):
```
jrs65, line 12: Your code seems to treat this as singleton, i.e. you have one master instance and it's keeping track of the different threads of profiling. Is there a reason not to make things easy and just create a new instance every time you want to track things?

caput/pipeline.py
```diff
@@ -593,11 +614,17 @@ def run(self):
...
+        Raises
+        ------
+        PipelineRuntimeError
+            If a task stage returns the wrong number of outputs.
         """
-
-        # Run the pipeline.
         while self.tasks:
             for task in list(self.tasks):  # Copy list so we can alter it.
+                if self._psutil_profiling:
```
jrs65, line 624: i.e. just create a new instance in here every time.

jrs65 (changes requested):

caput/misc/profiling.py
jrs65: A few more thoughts about this. I wrote a much more limited version of this for getting the amount of IO done in my branch `caput.misc.IOUsage`, and I think yours will basically be able to replace it, but there's maybe a few nice things it could do to fit in better in my code:
- Allow it to run as a context manager
- Accept a logger object rather than just using the module level logger

caput/pipeline.py
jrs65: I don't think running psutil has much overhead at all. I would be very tempted to run it *every* time (needing `psutil` to be an actual dependency), and have the only thing that changes be whether it actually outputs the stats.

caput/misc/profiling.py
jrs65: Also, another thought. You wrap the results in a namedtuple, I wonder if it would be better to do it with a dict. I find it useful to iterate over the items without knowing what they are, and you can't do that with a namedtuple without using semi-private attributes (either `._as_dict()` or `._fields`)

jrs65: Anyway, this all looks great! Thanks.

caput/misc/profiling.py
nritsche: Even before this addition, I think the caput package structure should rather get sorted into fewer sub-packages instead of adding more modules.

jrs65: Do you have a proposal as to how you would organise it? My initial reaction is that it only makes sense if you can group related things together and I really don't see what they'd be. At a module level things are currently logically grouped together which seems reasonable. 

I really don't see how `caput.misc.profiling` makes much sense though. It's not like anything else is more organised. I think we should probably just put this `caput.profiling` and consider a reorganisation later on, although backwards compatibility will make it difficult to do anything too major without lots of work.

nritsche: > Do you have a proposal as to how you would organise it? My initial reaction is that it only makes sense if you can group related things together and I really don't see what they'd be. At a module level things are currently logically grouped together which seems reasonable.
Right, but we can make subpackages out of most modules and split those into multiple files.

jrs65: Yeah, I'd agree with that for the big ones like `memh5` and `pipeline`, but I'm not sure if any of the others are large enough. I'd still put the profiling stuff in its own root-level module.

nritsche: I don't think `profiling` or `cache` are relevant enough to be a top level submodule.

anjakefala:
This work has been moved to this PR: https://github.com/radiocosmology/caput/pull/180

PR #168, opened 2021-03-31
feat(runner): add modules conf variable, specify module environment of a slurm pipeline job

anjakefala:
Closes #83


anjakefala:
Note, sticky modules like `StdEnv/2020 ` will not get purged with the current code. You would have to do `module purge --force`.

jrs65:
I don't think there's any reason to limit this to only running on slurm.

caput/scripts/runner.py
```diff
@@ -308,6 +315,16 @@ def queue(configfile, submit=False, lint=True):
     if sfile != dfile:
         shutil.copyfile(sfile, dfile)
 
+    if "modules" in rconf and rconf["modules"]:
+        modules = rconf["modules"]
+        modules = (modules,) if isinstance(modules, str) else modules
+        modstr = "module purge\nmodule load "
+        modstr += "\nmodule load ".join(modules)
```
anjakefala, line 322: Note that if the module does not exist, the pipeline job will continue to try to run. `module load` does not return an error code.

jrs65 (changes requested):
Can you make sure that the `STATUS` file gets set to `CRASHED` when the modules aren't available? I might be mistaken, but it doesn't seem like this version does that.

You might want to use some version of `trap <handler> ERR` to do that and then you can refactor the existing code at the end which sets `CRASHED`

jrs65 (approved):

anjakefala:
I am waiting on the results of two test runs. if they pass, I am going to go ahead and merge this. Wait time is high today for Cedar T____T

PR #169, opened 2021-04-01
feat(memh5): zarr support

nritsche:
TODO:
- [x] _distributed_group_to_zarr
- [x] unit tests
- [ ] test: format some old hdf5 files in zarr
- [x] run daily processing pipeline with zarr
- [x] we are creating `<filename>.zarr.sync` folders for zarr's synchronizer, but they don't get deleted by zarr. do we have to do that? https://github.com/harlowja/fasteners/issues/26
- [x] optimize striping for small (chunk) files: https://scicomp.aalto.fi/triton/usage/lustre/
- [x] increase chunk size to ~16MB (compressed on disk)

Also see https://github.com/radiocosmology/draco/pull/130
and chime-experiment/ch_pipeline#82

Closes https://github.com/chime-experiment/Pipeline/issues/33

**Edit by @anjakefala**
As of 2021-08-23, here is the discussed remaining work for this PR, along with any unchecked boxes above:
- [x] remove references to `chunkify()`
- [x] Ignore compression and chunking options when saving to `.h5` (log `DEBUG` level)
- [ ] add `zarr` to the base chime environment on Cedar, before we set the pipeline running with this new code
- [x] make sure the docs can build and upload to readthedocs; one way is to either make `bitshuffle` an optional dependency for building the docs, or build the docs and upload them through readthedocs through github actions (see ch_util)
      - Update: ch_util actually pushes the docs to github pages, not readthedocs: https://github.com/chime-experiment/ch_util/pull/10
- [x] clean up commits made by @anjakefala 
- [ ] @jrs65 and @anjakefala review the full PR
- [x] get a successful run with `test_daily` and then get a successful run with the full pipeline.

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -482,6 +482,62 @@ def logging_config(default=None):
     return prop
 
 
+def file_format(default=None):
+    """A property type that accepts only "zarr", or "hdf5".
+
+    Returns the selected `caput.fileformat.FileFormat` subclass or `caput.fileformats.HDF5` if `value == default`.
+
+    Parameters
+    ----------
+    default : optional
+        The optional default value.
+
+    Returns
+    -------
+    prop : Property
+        A property instance setup to validate a file format.
+
+    Raises
+    ------
+    ValueError
+        If the default value is not `"hdf5"` or `"zarr"`.
+
+    Examples
+    --------
+    Should be used like::
+
+        class Project:
+
+            mode = file_format(default='zarr')
+    """
+    options = ("hdf5", "zarr")
+
+    def _prop(val):
+        from . import fileformats
+
+        if val is None:
+            return None
+
+        if not isinstance(val, str):
+            CaputConfigError(
+                f"Input {repr(val)} is of type {type(val).__name__} (expected str or None)."
+            )
+
+        val = val.lower()
+        if val == "hdf5":
+            return fileformats.HDF5
+        if val == "zarr":
+            return fileformats.Zarr
+        raise CaputConfigError(f"Input {repr(val)} needs to be one of {options})")
+
+    if default is not None and (
+        (not isinstance(default, str)) or (default.lower() not in options)
+    ):
+        raise CaputConfigError(f"Default value {repr(default)} must be in {options}")
+
+    return Property(proptype=_prop, default=default)
+
+
 class _line_dict(dict):
     """A private dict subclass that also stores line numbers for debugging."""
 
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -60,6 +60,7 @@ import logging
 import numpy as np
 import h5py
 
+from . import fileformats
 from . import mpiutil
 from . import mpiarray
 from . import misc
@@ -67,6 +68,13 @@ from . import misc
 
 logger = logging.getLogger(__name__)
 
+try:
+    import zarr
+except ImportError as err:
+    logger.info(f"zarr support disabled. Install zarr to change this: {err}")
+    zarr_available = False
+else:
+    zarr_available = True
 
 # Basic Classes
 # -------------
@@ -390,7 +398,7 @@ class MemGroup(_BaseGroup):
         """Create a new instance by deep copying an existing group.
 
         Agnostic as to whether the group to be copied is a `MemGroup` or an
-        `h5py.Group` (which includes `hdf5.File` objects).
+        `h5py.Group` (which includes `h5py.File` and `zarr.File` objects).
 
         """
 
@@ -398,8 +406,14 @@ class MemGroup(_BaseGroup):
             self = cls()
             deep_group_copy(group, self)
             return self
+        elif isinstance(group, (str, bytes)):
+            file_format = fileformats.guess_file_format(group)
+            return cls.from_file(group, file_format=file_format)
         else:
-            return cls.from_hdf5(group)
+            raise RuntimeError(
+                f"Can't create an instance from type {type(group).__name__} "
+                f"(expected MemGroup, str or bytes)."
+            )
 
     @classmethod
     def from_hdf5(
@@ -411,7 +425,7 @@ class MemGroup(_BaseGroup):
         selections=None,
         convert_dataset_strings=False,
         convert_attribute_strings=True,
-        **kwargs
+        **kwargs,
     ):
         """Create a new instance by copying from an hdf5 group.
 
@@ -443,6 +457,63 @@ class MemGroup(_BaseGroup):
             Root group of loaded file.
         """
 
+        return cls.from_file(
+            filename,
+            distributed,
+            hints,
+            comm,
+            selections,
+            convert_dataset_strings,
+            convert_attribute_strings,
+            file_format=fileformats.HDF5,
+            **kwargs,
+        )
+
+    @classmethod
+    def from_file(
+        cls,
+        filename,
+        distributed=False,
+        hints=True,
+        comm=None,
+        selections=None,
+        convert_dataset_strings=False,
+        convert_attribute_strings=True,
+        file_format=fileformats.HDF5,
+        **kwargs,
+    ):
+        """Create a new instance by copying from a file group.
+
+        Any keyword arguments are passed on to the constructor for `h5py.File` or `zarr.File`.
+
+        Parameters
+        ----------
+        filename : string
+            Name of file to load.
+        distributed : boolean, optional
+            Whether to load file in distributed mode.
+        hints : boolean, optional
+            If in distributed mode use hints to determine whether datasets are
+            distributed or not.
+        comm : MPI.Comm, optional
+            MPI communicator to distributed over. If :obj:`None` use
+            :obj:`MPI.COMM_WORLD`.
+        selections : dict
+            If this is not None, it should map dataset names to axis selections as valid
+            numpy indexes.
+        convert_attribute_strings : bool, optional
+            Try and convert attribute string types to unicode. Default is `True`.
+        convert_dataset_strings : bool, optional
+            Try and convert dataset string types to unicode. Default is `False`.
+        file_format : `fileformats.FileFormat`
+            File format to use. Default `fileformats.HDF5`.
+
+        Returns
+        -------
+        group : memh5.Group
+            Root group of loaded file.
+        """
+
         if comm is None:
             comm = mpiutil.world
 
@@ -456,7 +527,7 @@ class MemGroup(_BaseGroup):
 
         if not distributed or not hints:
             kwargs["mode"] = "r"
-            with h5py.File(filename, **kwargs) as f:
+            with file_format.open(filename, **kwargs) as f:
                 self = cls(distributed=distributed, comm=comm)
                 deep_group_copy(
                     f,
@@ -464,15 +535,17 @@ class MemGroup(_BaseGroup):
                     selections=selections,
                     convert_attribute_strings=convert_attribute_strings,
                     convert_dataset_strings=convert_dataset_strings,
+                    file_format=file_format,
                 )
         else:
-            self = _distributed_group_from_hdf5(
+            self = _distributed_group_from_file(
                 filename,
                 comm=comm,
                 hints=hints,
                 selections=selections,
                 convert_attribute_strings=convert_attribute_strings,
                 convert_dataset_strings=convert_dataset_strings,
+                file_format=file_format,
             )
 
         return self
@@ -484,7 +557,7 @@ class MemGroup(_BaseGroup):
         hints=True,
         convert_attribute_strings=True,
         convert_dataset_strings=False,
-        **kwargs
+        **kwargs,
     ):
         """Replicate object on disk in an hdf5 file.
 
@@ -502,17 +575,58 @@ class MemGroup(_BaseGroup):
             understands. Default is `True`.
         convert_dataset_strings : bool, optional
             Try and convert dataset string types to bytestrings. Default is `False`.
+
         """
 
+        self.to_file(
+            filename,
+            mode,
+            hints,
+            convert_attribute_strings,
+            convert_dataset_strings,
+            fileformats.HDF5,
+            **kwargs,
+        )
+
+    def to_file(
+        self,
+        filename,
+        mode="w",
+        hints=True,
+        convert_attribute_strings=True,
+        convert_dataset_strings=False,
+        file_format=fileformats.HDF5,
+        **kwargs,
+    ):
+        """Replicate object on disk in an hdf5 or zarr file.
+
+        Any keyword arguments are passed on to the constructor for `h5py.File` or `zarr.File`.
+
+        Parameters
+        ----------
+        filename : str
+            File to save into.
+        hints : boolean, optional
+            Whether to write hints into the file that described whether datasets
+            are distributed, or not.
+        convert_attribute_strings : bool, optional
+            Try and convert attribute string types to a unicode type that HDF5
+            understands. Default is `True`.
+        convert_dataset_strings : bool, optional
+            Try and convert dataset string types to bytestrings. Default is `False`.
+        file_format : `fileformats.FileFormat`
+            File format to use. Default `fileformats.HDF5`.
+        """
         if not self.distributed:
-            with h5py.File(filename, mode, **kwargs) as f:
+            with file_format.open(filename, mode, **kwargs) as f:
                 deep_group_copy(
                     self,
                     f,
                     convert_attribute_strings=convert_attribute_strings,
                     convert_dataset_strings=convert_dataset_strings,
+                    file_format=file_format,
                 )
-        else:
+        elif file_format == fileformats.HDF5:
             if h5py.get_config().mpi:
                 _distributed_group_to_hdf5_parallel(
                     self,
@@ -529,6 +643,14 @@ class MemGroup(_BaseGroup):
                     convert_attribute_strings=convert_attribute_strings,
                     convert_dataset_strings=convert_dataset_strings,
                 )
+        else:
+            _distributed_group_to_zarr(
+                self,
+                filename,
+                mode,
+                convert_attribute_strings=convert_attribute_strings,
+                convert_dataset_strings=convert_dataset_strings,
+            )
 
     def create_group(self, name):
         """Create a group within the storage tree."""
@@ -574,7 +696,7 @@ class MemGroup(_BaseGroup):
         chunks=None,
         compression=None,
         compression_opts=None,
-        **kwargs
+        **kwargs,
     ):
         """Create a new dataset.
 
@@ -593,6 +715,11 @@ class MemGroup(_BaseGroup):
         distributed_axis : int, optional
             Axis to distribute the data over. If specified with initialisation
             data this will cause create a copy with the correct distribution.
+        compression : str or int
+            Name or identifier of HDF5 or Zarr compression filter.
+        compression_opts
+            See HDF5 and Zarr documentation for compression filters.
+            Compression options for the dataset.
 
         Returns
         -------
@@ -895,22 +1022,48 @@ class MemDataset(_MemObjMixin):
 
     @property
     def shape(self):
+        """
+        Shape of the dataset.
+
+        Not implemented in base class.
+        """
         raise NotImplementedError("Not implemented in base class.")
 
     @property
     def dtype(self):
+        """
+        numpy data type of the dataset.
+
+        Not implemented in base class.
+        """
         raise NotImplementedError("Not implemented in base class.")
 
     @property
     def chunks(self):
+        """
+        Chunk shape of the dataset.
+
+        Not implemented in base class.
+        """
         raise NotImplementedError("Not implemented in base class.")
 
     @property
     def compression(self):
+        """
+        Name or identifier of HDF5 compression filter for the dataset.
+
+        Not implemented in base class.
+        """
         raise NotImplementedError("Not implemented in base class.")
 
     @property
     def compression_opts(self):
+        """
+        Compression options for the dataset.
+
+        See HDF5 documentation for compression filters.
+        Not implemented in base class.
+        """
         raise NotImplementedError("Not implemented in base class.")
 
     def __getitem__(self, obj):
@@ -951,7 +1104,7 @@ class MemDatasetCommon(MemDataset):
         chunks=None,
         compression=None,
         compression_opts=None,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(**kwargs)
 
@@ -970,6 +1123,11 @@ class MemDatasetCommon(MemDataset):
         ----------
         data : np.ndarray
             Array to initialise from.
+        compression : str or int
+            Name or identifier of HDF5 or Zarr compression filter.
+        compression_opts
+            See HDF5 and Zarr documentation for compression filters.
+            Compression options for the dataset.
 
         Returns
         -------
@@ -1024,7 +1182,17 @@ class MemDatasetCommon(MemDataset):
 
     @chunks.setter
     def chunks(self, val):
-        self._chunks = val
+        if val is None:
+            chunks = val
+        elif len(val) != len(self.shape):
+            raise ValueError(
+                f"Chunk size {val} is not compatible with dataset shape {self.shape}."
+            )
+        else:
+            chunks = ()
+            for i, l in enumerate(self.shape):
+                chunks += (min(val[i], l),)
+        self._chunks = chunks
 
     @property
     def compression(self):
@@ -1105,7 +1273,7 @@ class MemDatasetDistributed(MemDataset):
         chunks=None,
         compression=None,
         compression_opts=None,
-        **kwargs
+        **kwargs,
     ):
         super().__init__(**kwargs)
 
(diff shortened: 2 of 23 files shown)


caput/mpiarray.py
```diff
@@ -830,7 +882,7 @@ def to_file(
                 islice, fslice = _partition_sel(
                     sel, split_axis, self.global_shape[split_axis], part
                 )
-                dset[islice] = self[fslice]
+                dset[islice] = self.local_array[fslice]
```
anjakefala, line 885: yay

nritsche:
https://github.com/chime-experiment/ch_pipeline/pull/82

anjakefala:
By default, `mkchimeenv.sh` currently does not install `zarr` when it installs `caput`. We should perhaps add `zarr` to our chime environment as part of shipping this.

caput/memh5.py
```diff
@@ -411,7 +425,7 @@ def from_hdf5(
         selections=None,
         convert_dataset_strings=False,
         convert_attribute_strings=True,
-        **kwargs
+        **_,
```
anjakefala, line 428: Rick is not passing on the `kwargs` here, I am not sure why, yet. Open question if this that we want to do?

anjakefala: The more I look at this, the more I think that we do want to pass on the `kwargs`, and removing them was possibly a linting error.

caput/memh5.py
```diff
@@ -484,7 +556,9 @@ def to_hdf5(
         hints=True,
         convert_attribute_strings=True,
         convert_dataset_strings=False,
-        **kwargs
+        compression=None,
+        compression_opts=None,
+        **kwargs,
```
anjakefala, line 561: We may want to add `chunks` as an explicit argument.

caput/memh5.py
```diff
@@ -503,16 +577,61 @@ def to_hdf5(
         convert_dataset_strings : bool, optional
             Try and convert dataset string types to bytestrings. Default is `False`.
```
anjakefala, line 578: And add the `compression` options here in the docstrings.

caput/memh5.py
anjakefala: Oh, chunking does not happen with `MemGroup`?

caput/memh5.py
```diff
@@ -1282,9 +1450,11 @@ class MemDiskGroup(_BaseGroup):
     detect_subclass: boolean, optional
         If *data_group* is specified, whether to inspect for a
         '__memh5_subclass' attribute which specifies a subclass to return.
+    file_format : `fileformats.FileFormat`
+        File format to use. File format will be guessed if not supplied. Default `None`.
     """
 
-    def __init__(self, data_group=None, distributed=False, comm=None):
+    def __init__(self, data_group=None, distributed=False, comm=None, file_format=None):
```
anjakefala, line 1457: I think `file_format` can have a fallback value of `None` in a couple of places, and I am wondering if this ever breaks anything. Testing a form of `.yaml` where you have a `.zarr` filename, but do not specify the `output_format`.

caput/memh5.py
```diff
@@ -2066,28 +2295,28 @@ def get_h5py_File(f, **kwargs):
...
+        if file_format == fileformats.Zarr and not zarr_available:
+            raise RuntimeError("Unable to open zarr file. Please install zarr.")
         try:
-            f = h5py.File(f, **kwargs)
+            f = file_format.open(f, **kwargs)
         except IOError as e:
             msg = "Opening file %s caused an error: " % str(f)
             raise IOError(msg + str(e)) from e
-    return f, opened
+        return f, file_format == fileformats.HDF5
```
anjakefala, line 2309: Why does it not return `True` for `zarr` as well? Are `zarr` files by default "already opened"?

caput/memh5.py
```diff
@@ -2226,27 +2470,40 @@ def deep_group_copy(
             if convert_dataset_strings:
                 # Convert unicode strings back into ascii byte strings. This will break
                 # if there are characters outside of the ascii range
-                if isinstance(g2, h5py.Group):
+                if isinstance(g2, file_format.module.Group):
```
anjakefala, line 2473: This will `Exception` if `not zarr_available`

caput/memh5.py
```diff
@@ -2226,27 +2470,40 @@ def deep_group_copy(
...
+                if compression_entry is not None:
+                    compression = compression_entry
+                    compression_opts = getattr(entry, "compression_opts", None)
+                compression_kwargs = file_format.compression_kwargs(
+                    compression=compression,
+                    compression_opts=compression_opts,
+                    compressor=getattr(entry, "compressor", None),
+                )
+            else:
+                # use HDF5 compression args format
```
anjakefala, line 2496: If this codeblock is going to run for `file_format.Zarr` as well, then we cannot have `HDF5.compression_kwargs` return `{}` by default. I actually do not understand what the `if isinstance(g2, file_format.module.Group)` check is about. Is this, if it is a `dataset`, then we use `HDF5.compression_kwargs`? Does `zarr` have datasets?

caput/memh5.py
```diff
@@ -2471,9 +2727,12 @@ def _copy_to_file(memgroup, h5group):
...
                         dtype=data.dtype,
-                        chunks=item.chunks,
-                        compression=item.compression,
-                        compression_opts=item.compression_opts,
+                        chunks=item.chunks
+                        if fileformats.HDF5.compression_enabled()
+                        else None,
+                        **fileformats.HDF5.compression_kwargs(
+                            item.compression, item.compression_opts
+                        ),
```
anjakefala, line 2735: I wonder why `compression_kwargs` are used here, but not at 2709..

caput/memh5.py
```diff
@@ -2503,13 +2762,135 @@ def _copy_to_file(memgroup, h5group):
...
+                        data = check_unicode(item)
+
+                    # Write common data from rank 0
+                    if memgroup.comm.rank == 0:
+                        dset = group.create_dataset(
+                            key,
+                            shape=data.shape,
+                            dtype=data.dtype,
+                            chunks=item.chunks,
+                            **fileformats.Zarr.compression_kwargs(
```
anjakefala, line 2846: I would say, we should ensure every `create_dataset` has the `fileformats.compression_kwargs` treatment...

caput/memh5.py
```diff
@@ -2503,13 +2762,135 @@ def _copy_to_file(memgroup, h5group):
...
+            f.attrs["__memh5_distributed_file"] = True
+
+    # Final synchronisation
+    group.comm.Barrier()
+
+
+def _distributed_group_from_file(
+    fname,
+    comm=None,
+    _=True,
```
anjakefala, line 2889: `_=True` used to be `hints=True`. I am guessing `hints` is deprecated?

caput/memh5.py
```diff
@@ -2503,13 +2762,135 @@ def _copy_to_file(memgroup, h5group):
...
+    convert_dataset_strings=False,
+    convert_attribute_strings=True,
+    file_format=fileformats.HDF5,
+    **kwargs,
 ):
     """
     Restore full tree from an HDF5 file into a distributed memh5 object.
```
anjakefala, line 2896: "from an HDF5 file or `Zarr group`"

caput/memh5.py
```diff
@@ -2594,10 +2980,14 @@ def _copy_from_file(h5group, memgroup, selections=None):
...
-    with misc.open_h5py_mpi(fname, "r", comm=comm) as f:
+    if file_format == fileformats.HDF5:
+        # Open file on all ranks
+        with misc.open_h5py_mpi(fname, "r", comm=comm) as f:
 
-        # Start recursive file read
+            # Start recursive file read
+            _copy_from_file(f, group, selections)
+    else:
+        f = file_format.open(fname, "r")
```
anjakefala, line 2990: From this I am assuming that this is intended to work with `Zarr` as well

caput/mpiarray.py
```diff
@@ -631,12 +631,61 @@ def from_hdf5(cls, f, dataset, comm=None, axis=0, sel=None):
...
-        use_mpi = axis > 0
+        return cls.from_file(f, dataset, comm, axis, sel, file_format=fileformats.HDF5)
 
-        # Read the file. Opening with MPI if requested, and we can
-        fh = misc.open_h5py_mpi(f, "r", use_mpi=use_mpi, comm=comm)
+    @classmethod
+    def from_file(
+        cls, f, dataset, comm=None, axis=0, sel=None, file_format=fileformats.HDF5
+    ):
+        """Read MPIArray from an HDF5 dataset in parallel.
```
anjakefala, line 640: "/Zarr Dataset"

caput/mpiarray.py
```diff
@@ -631,12 +631,61 @@ def from_hdf5(cls, f, dataset, comm=None, axis=0, sel=None):
...
+            # Blosc may share incorrect global state amongst processes causing programs to hang.
+            # See https://zarr.readthedocs.io/en/stable/tutorial.html#parallel-computing-and-synchronization
+            try:
+                import numcodecs
+            except ImportError:
+                raise RuntimeError("Install numcodecs to read from zarr files.")
+            numcodecs.blosc.use_threads = False
+
+            if isinstance(f, str):
+                fh = file_format.open(f, "r")
```
anjakefala, line 682: Okay, so we do not open `zarr` with MPI

caput/mpiarray.py
```diff
@@ -712,12 +770,11 @@ def to_hdf5(
 
         Parameters
         ----------
-        filename : str, h5py.File or h5py.Group
+        f : str, h5py.File or h5py.Group
             File to write dataset into.
         dataset : string
```
anjakefala, line 775: Add to docstring

caput/mpiarray.py
```diff
@@ -730,39 +787,38 @@ def to_hdf5(
...
         # Only use collective IO if:
         # - there are no null slices (h5py bug)
         # - we are not distributed over axis=0 as there is no advantage for
         #   collective IO which is usually slow
         # - unless we want to use compression/chunking
         # TODO: change if h5py bug fixed
+        # https://github.com/h5py/h5py/issues/965
         # TODO: better would be a test on contiguous IO size
         use_collective = (
             fh.is_mpi and no_null_slices and (self.axis > 0 or compression is not None)
```
anjakefala, line 805: On second thought, we may want to include here that compression is turned off for `HDF5`. Because since we are not doing compression, there is a possibility of using collective IO.

caput/mpiarray.py
```diff
@@ -730,39 +787,38 @@ def to_hdf5(
...
         )
 
-        if fh.is_mpi and not use_collective:
+        if (
+            fh.is_mpi
+            and (not use_collective)
+            or (not fileformats.HDF5.compression_enabled())
+        ):
             # Need to disable compression if we can't use collective IO
```
anjakefala, line 813: Though, I do not really understand collective IO, and its relationship to compression.

caput/mpiarray.py
```diff
@@ -730,39 +787,38 @@ def to_hdf5(
...
         # TODO: better would be a test on contiguous IO size
         use_collective = (
             fh.is_mpi and no_null_slices and (self.axis > 0 or compression is not None)
         )
 
-        if fh.is_mpi and not use_collective:
+        if (
+            fh.is_mpi
+            and (not use_collective)
+            or (not fileformats.HDF5.compression_enabled())
```
anjakefala, line 811: This is where I added the compression enabled check, to turn off `chunks`, and `compression`.

caput/mpiarray.py
```diff
@@ -730,39 +787,38 @@ def to_hdf5(
...
             # Need to disable compression if we can't use collective IO
             chunks, compression, compression_opts = None, None, None
 
-        dset = fh.create_dataset(
+        sel = self._make_selections()
+
+        # Split the axis to get the IO size under ~2GB (only if MPI-IO)
+        split_axis, partitions = self._partition_io(skip=(not fh.is_mpi))
+
+        fh.create_dataset(
```
anjakefala, line 821: If `create_dataset` is where Rick has been adding `fileformats.compression_kwargs`, then maybe the better place to turn off chunking et al is within `create_dataset` calls.

caput/mpiarray.py
```diff
@@ -772,18 +828,149 @@ def to_hdf5(
...
+        # Split the axis
+        split_axis, partitions = self._partition_io(skip=True)
+
+        if self.comm.rank == 0:
+            group.create_dataset(
+                dataset,
+                shape=self.global_shape,
+                dtype=self.dtype,
+                chunks=chunks,
+                **extra_args,
```
anjakefala, line 913: These are `**Zarr.compression_args()`

caput/pipeline.py
```diff
@@ -1263,7 +1266,13 @@ def read_process_write(self, input, input_filename, output_filename):
             output_dirname = os.path.dirname(output_filename)
             if not os.path.isdir(output_dirname):
                 os.makedirs(output_dirname)
-            self.write_output(output_filename, output)
+            self.write_output(
```
anjakefala, line 1269: `write_output()` lost its `self` parameter (It is now a `staticmethod`, so I am not quite sure how this has been working for my tests!

caput/pipeline.py
```diff
@@ -1436,34 +1449,51 @@ class H5IOMixin:
     # TODO, implement reading on disk (i.e. no copy to memory).
     # ondisk = config.Property(default=False, proptype=bool)
 
-    def read_input(self, filename):
+    @staticmethod
+    def read_input(filename):
```
anjakefala, line 1453: I wonder why Rick made all of these `staticmethod`.

caput/tests/conftest.py
```diff
@@ -89,10 +90,51 @@ def read_input(self, filename):
     def read_output(self, filename):
         raise NotImplementedError()
 
-    def write_output(self, filename, output):
+    def write_output(self, filename, output, file_format=None, **kwargs):
```
anjakefala, line 93: Do these need to be `staticmethod`?

caput/tod.py
```diff
@@ -118,7 +120,7 @@ def __init__(self, files):
 
         # Fetch all meta data.
         time = np.copy(data_empty.time)
-        first_file, toclose = memh5.get_h5py_File(files[0])
```
anjakefala, line 121: We went from a default of `HDF5` to a default of "guessing from filename". This will be fine if we use a consistent suffix with our hdf5 files.

setup.py
```diff
@@ -61,7 +61,15 @@
     """,
     python_requires=">=3.6",
     install_requires=requires,
-    extras_require={"mpi": ["mpi4py>=1.3"], "profiling": ["psutil", "pyinstrument"]},
+    extras_require={
+        "mpi": ["mpi4py>=1.3"],
+        "compression": [
+            "bitshuffle @ git+https://github.com/kiyo-masui/bitshuffle.git",
+            "numcodecs==0.7.3",
```
anjakefala, line 68: The `compression` extra is new, `bitshuffle` and `numcodecs` will not install by default.

caput/memh5.py
```diff
@@ -2276,7 +2533,7 @@ def _distributed_group_to_hdf5_serial(
     hints=True,
     convert_dataset_strings=False,
     convert_attribute_strings=True,
-    **kwargs
+    **kwargs,
```
anjakefala, line 2536: If I understand correctly, the `create_dataset` in `_distributed_group_to_hdf5_serial` is not going through `compression_kwargs` because the bug only pops up for parallel.

anjakefala: The compression and truncation bug for HDF5.

caput/memh5.py
anjakefala: You can remove the `output_format` and `.zarr` outputs are still accurately typed. Guessing this is fine, then.

caput/memh5.py
anjakefala: I removed the return `{}` by default in `HDF5.compression_kwargs` until we understand this better.

caput/memh5.py
anjakefala: It was confirmed by Richard that the bug only matters for parallelised IO, but we should still be formatting the `compression_kwargs`, I think, so I will add that.

caput/memh5.py
anjakefala: `Zarr` fileformats do not necessarily have a `close()` method; we should take this outside of `get_file`, and decide on how that file format gets closed.

caput/memh5.py
anjakefala: If `else`, that is the `in-memory` case, and by default Rick uses the HDF5 compression kwarg format for that case.

caput/memh5.py
anjakefala: Leave a comment that it does not do anything in this method.

caput/mpiarray.py
anjakefala: We should log that `compression` has been turned off, or throw an Exception.

caput/pipeline.py
anjakefala: Probably linter

caput/pipeline.py
```diff
@@ -1263,7 +1266,13 @@ def read_process_write(self, input, input_filename, output_filename):
             output_dirname = os.path.dirname(output_filename)
             if not os.path.isdir(output_dirname):
                 os.makedirs(output_dirname)
-            self.write_output(output_filename, output)
+            _OneAndOne.write_output(
```
anjakefala, line 1269: Revert this back to `self.write_output`, if `self.` will work with a `staticmethod`.

or `self.__class__.write_output`

caput/tod.py
anjakefala: If a guess is not conclusive, maybe hdf5?

setup.py
```diff
@@ -61,7 +61,15 @@
...
+        ],
+        "zarr": ["zarr==2.8.1"],
```
anjakefala, line 70: It does not necessarily make sense to separate `zarr` from `numcodecs`, since they are used together.

See if you can add platform specifiers, (Check for MACOS, and if MACOS, do not install `zarr`.

caput/pipeline.py
anjakefala: It does!

caput/memh5.py
```diff
@@ -2168,6 +2414,9 @@ def deep_group_copy(
     selections=None,
     convert_dataset_strings=False,
     convert_attribute_strings=True,
+    file_format=fileformats.HDF5,
+    compression=None,
+    compression_opts=None,
```
jrs65, line 2419: Should these be able to be specified per dataset?

caput/memh5.py
```diff
@@ -2330,21 +2602,34 @@ def _distributed_group_to_hdf5_serial(
...
+            else:
+                # disable compression if not enabled for HDF5 files
+                # https://github.com/chime-experiment/Pipeline/issues/33
+                arr.to_hdf5(
+                    fname,
+                    entry.name,
+                    chunks=None,
+                    compression=None,
+                    compression_opts=None,
+                )
```
jrs65, line 2632: It seems like this would be better structured with the conditional formatting the arguments and then a single call to `.to_hdf5`

jrs65: Or move the conditional into `.to_hdf5`

caput/memh5.py
```diff
@@ -2445,13 +2733,24 @@ def _copy_to_file(memgroup, h5group):
...
 
                     # Write to file from MPIArray
-                    data.to_hdf5(
-                        h5group,
-                        key,
-                        chunks=item.chunks,
-                        compression=item.compression,
-                        compression_opts=item.compression_opts,
-                    )
+                    if fileformats.HDF5.compression_enabled():
```
jrs65, line 2736: Same comments as above.

jrs65: Also, how did we avoid this before? I guess we just relied on the user not setting any compression as it would break.

caput/mpiarray.py
```diff
@@ -712,12 +770,11 @@ def to_hdf5(
 
         Parameters
         ----------
-        filename : str, h5py.File or h5py.Group
+        f : str, h5py.File or h5py.Group
             File to write dataset into.
         dataset : string
             Name of dataset to write into. Should not exist.
```
jrs65, line 776: Needs documentation for all parameters.

caput/pipeline.py
```diff
@@ -1330,6 +1340,9 @@ class SingleBase(_OneAndOne):
 
     input_filename = config.Property(default="", proptype=str)
     output_filename = config.Property(default="", proptype=str)
+    output_format = config.file_format()
+    output_compression = config.Property(default=None, proptype=str)
+    output_compression_opts = config.Property(default=None)
```
jrs65, line 1345: I think we need to review if these settings should be *per* dataset.

caput/tests/test_selection_parallel.py
```diff
@@ -2,59 +2,86 @@
...
 
     comm.Barrier()
 
     # tear down
-
     if comm.rank == 0:
-        file_names = glob.glob(fname + "*")
-        for fname in file_names:
-            os.remove(fname)
+        rm_all_files(zarr_file)
```
jrs65, line 59: It does seem like this could factor out the memh5 container creation

jrs65 (changes requested):

jrs65:
I added a bunch of comments, there not all that actionable, some are just notes to myself wondering how something is meant to work.

caput/tests/test_truncate.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
```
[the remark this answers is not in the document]
tristpinsm: yeah I guess it could be useful to test the behaviour of some edge cases, if only to be aware of their outcomes. These could include:
* precision set to 0/inf/NaN
* weight set to 0/inf/NaN
* value set to inf/NaN
* combinations of the above
* not an edge case exactly but truncating negative numbers

More generally a more robust test that I was using when developing these functions was to generate an ensemble of random numbers, truncate them, and check that the distribution of truncation errors was what you expect. That might be cumbersome and difficult to implement in this context though.

Another thing to note is there is a comment that claims one of the tests fails...

setup.py
jrs65: numcodecs should work on macos now.

setup.py
```diff
@@ -61,7 +61,15 @@
     """,
     python_requires=">=3.6",
     install_requires=requires,
-    extras_require={"mpi": ["mpi4py>=1.3"], "profiling": ["psutil", "pyinstrument"]},
+    extras_require={
+        "mpi": ["mpi4py>=1.3"],
+        "compression": [
+            "bitshuffle @ git+https://github.com/kiyo-masui/bitshuffle.git",
+            "numcodecs==0.7.3",
+            "zarr==2.8.1;platform_system!='Darwin'", # zarr is not supported on MacOS
```
jrs65, line 69: Can remove conditional here as zarr should work on macOS again.

caput/tests/test_truncate.py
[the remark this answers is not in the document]
jrs65: Let's add a few special cases in for this.

caput/memh5.py
```diff
@@ -502,17 +577,69 @@ def to_hdf5(
...
+        """
         if not self.distributed:
-            with h5py.File(filename, mode, **kwargs) as f:
+            with file_format.open(filename, mode, **kwargs) as f:
                 deep_group_copy(
                     self,
                     f,
                     convert_attribute_strings=convert_attribute_strings,
                     convert_dataset_strings=convert_dataset_strings,
+                    file_format=file_format,
```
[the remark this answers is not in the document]
jrs65: This seems to be very confusing. We should try to simplify.

caput/tests/test_selection_parallel.py
jrs65: A nicety, but not very important.

caput/mpiarray.py
anjakefala: Oh *wow*, they are not documented anywhere.

caput/config.py
```diff
@@ -510,7 +510,7 @@ class Project:
 
             mode = file_format(default='zarr')
     """
-    options = ("hdf5", "zarr", None)
```
tristpinsm: without a `None` default option, the `caput.fileformats.check_file_format` guessing functionality is broken.

caput/memh5.py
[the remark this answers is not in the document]
jrs65: Problem is with file_format, compression opts have been removed.

caput/mpiarray.py
jrs65: Still need doc entry for `chunks`

caput/tests/test_selection_parallel.py
jrs65: I can do this when I fix the tests.

caput/config.py
jrs65: @tristpinsm is looking into this.

jrs65 (approved):

anjakefala:
:clap: :clap: :clap: :clap:

jrs65:
> 👏 👏 👏 👏

Slowly getting there. Now I'm fighting to merge mpiarray in!!

PR #172, opened 2021-04-06
Allow variable substitution into pipeline scripts

jrs65:
This is a bit of a hack at the moment, but is a first attempt at some functionality I'd like to be in the pipeline. I've been using it to run grids of test jobs like this:
```bash
$ caput-pipeline template-run --var nside=128,256,512 --var redshift=10,30 testconv_za.yaml.tmpl
```

and the `.tmpl` file contains things like:
```yaml
    - type: draco.core.io.LoadBasicCont
      out: lss
      params:
        files:
          - lss_n{nside}.h5

    - type: cora.signal.lss.GenerateConstantBias
      in: lss
      out: biased
      params:
        bias: 0.0
        lightcone: false
        redshift: {redshift}
        output_name: bf_n{nside}_z{redshift}.h5
        save: true
```

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -93,7 +93,8 @@ class Property:
         # Ensure the property name has been found and set
         self._set_propname(obj)
 
-        # If the value has not been set, return the default, otherwise return the actual value.
+        # If the value has not been set, return the default, otherwise return the
+        # actual value.
         if self.propname not in obj.__dict__:
             return self.proptype(self.default) if self.default is not None else None
         return obj.__dict__[self.propname]
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -1,7 +1,8 @@
+import itertools
 import os
 import sys
-
 from pathlib import Path
+import tempfile
 
 import click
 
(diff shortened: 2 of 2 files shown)


jrs65:
@anjakefala @nritsche this isn't a finished thing, but I wanted to get your thoughts on the right way of inserting functionality like this into the pipeline running.

jrs65:
Ugh. Does the linter really not allow TODO items?

nritsche (approved):
Makes sense.

How often did you merge a TODO comment and fixed it later?

anjakefala:
I see now with why expanded templating is so important!

Some stray thoughts I had:
- That outer join is definitely the hackiest part, and it would be good for us to think about how we could make that a little cleaner. If it has been working for you, it is probably fine for now. 
- This means that the user should supply exactly one `--var` for every `{}` in the template file, right? It would be nice if we could add additional error checking / communication about that. The logging of the `varstring` is a good touch.
- I played with splitting based on a regex string, which would allow us to escape commas, but the result has been clunky =( This might require an actual tokenizing function that we write.

```
import re
hi = 'hello,my\,name,is'
>>> re.split(r'(?<!\\),', hi)
['hello', 'my\\,name', 'is']
>>> re.split(r'(?<!\\),', hi)[1][3]
','
>>> re.split(r'(?<!\\),', hi)[1][2]
'\\'
```

anjakefala:
Just so I understand the varstring at the end

```
$ caput-pipeline template-run --var nside=128,256,512 --var redshift=10,30 testconv_za.yaml.tmpl
```

Will result with

```
[ {"nside": 128,
    "redshift": 10 },
 {"nside": 128,
    "redshift: 30},
   {"nside": 256,
    "redshift": 10},
    {"nside": 256,
    "redshift": 30},
    {"nside": 512, 
    "redshift": 10},
    {"nside: 512,
    "redshift": 30}
]
```

Is that correct?

jrs65:
@anjakefala @nritsche I think it's easier just to get this merged now. I've reworked the generation of the variable sets by using `itertools.product` which is a little cleaner, if you're able to have another looks.

nritsche:
The `memh5` changes in the force push seem unrelated. What are they about?

jrs65:
> The `memh5` changes in the force push seem unrelated. What are they about?

I don't think there actually are any, it's just an issue with how it calculated the diff.

jrs65:
To confuse you more I just force pushed again to update against the latest master.

nritsche (approved):

jrs65:
Thanks Rick!

PR #174, opened 2021-04-13
build: update versioneer

anjakefala:
- we need versioneer at least 0.19 to have get_cmdclass accept an argument

See: https://github.com/python-versioneer/python-versioneer/issues/161#issuecomment-693070731


caput/_version.py
```diff
@@ -116,7 +116,7 @@ def versions_from_parentdir(parentdir_prefix, root, verbose):
     """
     rootdirs = []
 
-    for _ in range(3):
```
nritsche, line 119: `i` is still not used here, so this will trigger some linter warnings.

caput/_version.py
```diff
@@ -241,7 +241,7 @@ def git_pieces_from_vcs(tag_prefix, root, verbose, run_command=run_command):
     if sys.platform == "win32":
         GITS = ["git.cmd", "git.exe"]
 
-    _, rc = run_command(GITS, ["rev-parse", "--git-dir"], cwd=root, hide_stderr=True)
```
nritsche, line 244: Same as above

caput/_version.py
```diff
@@ -524,7 +524,7 @@ def get_versions():
         # versionfile_source is the relative path from the top of the source
         # tree (where the .git directory might live) to this file. Invert
         # this to find the root from __file__.
-        for _ in cfg.versionfile_source.split("/"):
```
nritsche, line 527: Same as above

nritsche (changes requested):

nritsche (approved):
:+1:

caput/_version.py
anjakefala: I saw that! I want to ignore the linter warnings for this file, but would you prefer for me to modify `_version.py`?

PR #175, opened 2021-04-28
feat(truncate): a new module for precision truncation of data

jrs65:


jrs65:
@nritsche I think this is good enough for someone to start looking at. I might still make some changes to the truncate code but that can come in another PR.

caput/misc.py
```diff
@@ -1,11 +1,15 @@
 """A set of miscellaneous routines that don't really fit anywhere more specific."""
```
nritsche, line 1: Can you review https://github.com/radiocosmology/caput/pull/164 so we can merge that before making changes to `caput.misc`?

caput/misc.py
```diff
@@ -357,3 +361,161 @@ class : class object
...
+        if psutil.MACOS:
+            d = psutil.disk_io_counters()
+        else:
+            p = psutil.Process()
+            d = p.io_counters()
+
+        # There doesn't seem to be a public API for this
+        return d._asdict()
+
+    def _units(self, key):  # pylint: disable=no-self-use
```
nritsche, line 486: Why not make this a `@staticmethod`?

caput/scripts/runner.py
```diff
@@ -123,43 +124,18 @@ def run(configfile, loglevel, profile, profiler):
...
-        rank = mpiutil.rank
-
-        if profiler == "cProfile":
-            pr.disable()
-            pr.dump_stats("profile_%i.prof" % mpiutil.rank)
-        elif profiler == "pyinstrument":
-            pr.stop()
-            with open("profile_%i.txt" % rank, "w") as fh:
-                fh.write(pr.output_text(unicode=True))
+    with Profiler(profile, profiler=profiler):
```
nritsche, line 127: How would `IOUsage` be used here? Maybe I missed sth

jrs65: Yeah, it doesn't get used there. I just have a few blocks in driftscan that I want to see how much IO is done and so I wrap them in a block like:
```python
with IOUsage(logger=logger):
    bt.generate_mfiles()
```

caput/misc.py
jrs65: That sounds reasonable I guess. Although I'm pretty sure I made @anjakefala to use functions rather than staticmethods. In this case I think a staticmethod is justifiable as these really should be private and specific to this class.

jrs65:
Truncate code from: `https://github.com/radiocosmology/draco/blob/master/draco/util/truncate.pyx`

caput/mpiutil.py
```diff
@@ -677,6 +679,60 @@ def parallel_rows_write_hdf5(fname, dsetname, local_data, shape, comm=_comm):
     lock_and_write_buffer(local_data, fname, offset + sr * nc, lr * nc)
 
 
+# Need to disable the pylint error here as we only need to implement one method
+class MPILogFilter(logging.Filter):  # pylint: disable=too-few-public-methods
```
nritsche, line 683: Will this replace the one in draco?

caput/truncate.pyx
```diff
@@ -0,0 +1,161 @@
+"""Routines for truncating data to a specified precision."""
```
nritsche, line 1: Where is this used?

caput/mpiutil.py
jrs65: Yup. When I've got this merged I'll remove it from draco.

caput/truncate.pyx
jrs65: Both in draco (in the future) and in driftscan (currently).

jrs65:
Okay. This C++ build failure is the last straw. I'm disabling LGTM until it's not a piece of junk.

nritsche (approved):

jrs65:
Great. Thanks @nritsche !

PR #176, opened 2021-05-10
feat(memh5): add a __memh5_distributed_axis hint

ssiegelx:
If hints are enabled, then the axis that each dataset is
distributed over at the time it is written to disk will be
saved to that dataset's attributes.  The dataset will then
be distributed over that axis the next time it is read from disk.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2397,6 +2397,9 @@ def _distributed_group_to_hdf5_serial(
 
                     if hints:
                         f[entry.name].attrs["__memh5_distributed_dset"] = True
+                        f[entry.name].attrs[
+                            "__memh5_distributed_axis"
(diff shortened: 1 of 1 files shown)


ssiegelx:
I have been getting memory errors when reading `RingMaps` from disk because the first axis of the `map` dataset is `beam`, which is length 1.  Since `memh5` always distributes over the first axis when reading from disk, each node attempts to read the entire `map` dataset.

This fixes the problem by adding a new hint to the attributes of each dataset that indicates what axis the dataset was distributed over when it was written to disk.

jrs65 (approved):
Great. Thanks Seth! I've had these problems before too, and thought I'd fixed them somewhere, but I think that branch is lost to the mists of time!

PR #182, opened 2021-07-13
build: build using the oldest supported numpy version

anjakefala:
Binaries compiled with old Numpy versions are ABI compatible with
newer Numpy versions, but not vice-versa.

To avoid running issues into on cedar, we should build using the oldest
compatible numpy version. We can then run packages using whatever newer
Numpy version we want.

Conversation here: https://github.com/radiocosmology/draco/pull/145


jrs65 (approved):

jrs65:
You might consider removing the ignore file.

PR #184, opened 2021-07-21
Useful enhancements for running the pipeline

jrs65:
- fix(config): format string bugs
- feat(template-run): add option to submit templat jobs to the queue
- feat(runner,mpiutil): allow control of MPI abort on exception

### DIFF
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -153,8 +153,15 @@ def run(configfile, loglevel, profile, profiler):
     ),
     metavar="<NAME>=<VALUE>[,<VALUE2>...]",
 )
-def template_run(templatefile, var):
-    """Run a pipeline immediately from the given TEMPLATEFILE.
+@click.option(
+    "--submit/--nosubmit", default=False, help="Submit into the batch queue"
+)
+@click.pass_context
+def template_run(ctx, templatefile, submit, var):
+    """Run a pipeline from the given TEMPLATEFILE.
+
+    This is either run immediately (default), or can be placed in the batch
+    queue with the --submit flag.
 
     Template variable substitutions are specified with `--var <varname>=<val>`
     arguments, with one for each variable. `<val>` may be a comma separated list, in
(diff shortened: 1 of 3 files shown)


caput/scripts/runner.py
```diff
@@ -153,7 +153,11 @@ def run(configfile, loglevel, profile, profiler):
     ),
     metavar="<NAME>=<VALUE>[,<VALUE2>...]",
 )
-def template_run(templatefile, var):
+@click.option(
+    "--submit/--nosubmit", default=False, help="Submit into the batch queue"
+)
+@click.pass_context
+def template_run(ctx, templatefile, submit, var):
     """Run a pipeline immediately from the given TEMPLATEFILE.
```
anjakefala, line 172: > From the given TEMPLATEFILE, run a pipeline immediately (default), or submit it into the batch queue (`--nosubmit`).

anjakefala: Docstring rewording suggestion ^

caput/scripts/runner.py
```diff
@@ -190,8 +194,11 @@ def template_run(templatefile, var):
             tfh.write(expanded_string)
             tfh.flush()
 
-            # Run the pipeline
-            Manager.from_yaml_file(tfh.name).run()
+            # Run or queue the pipeline
+            if queue:
+                ctx.invoke(queue, configfile=tfh.name)
```
anjakefala, line 210: Do you need to pass `submit=True`?

caput/scripts/runner.py
```diff
@@ -190,8 +194,11 @@ def template_run(templatefile, var):
             tfh.write(expanded_string)
             tfh.flush()
 
-            # Run the pipeline
-            Manager.from_yaml_file(tfh.name).run()
+            # Run or queue the pipeline
+            if queue:
```
anjakefala, line 209: Where does `queue` come from?

anjakefala:
I have some questions about 5f7f8c8 that might be the result of my relatively poor understanding of `click` and contexts.

Everything else looks great! I also do not think it will conflict too badly with my PR. Thanks for fixing the format strings

caput/scripts/runner.py
jrs65: Haha. Good spot. I already fixed that one but I haven't pushed the commit. I found it when it submitted a bunch of jobs into the queue that I had wanted it to run immediately!

caput/scripts/runner.py
jrs65: Yeah, the issue is that it should be `if submit` !

caput/scripts/runner.py
anjakefala: Ah! Okay, once that is pushed, we can merge this in. =)

anjakefala (approved):

PR #185, opened 2021-07-30
feat(runner): add psutil based profiling to caput-pipeline

anjakefala:
To turn on profiling with psutil, pass the flag `--psutil` with pipeline `queue` or `run`.

The profiling results will be saved to csv files, one for each rank, in the pipeline's `work_dir` as `perf_{rank}.csv`.


jrs65:
Great. Thanks!

Issue #186, opened 2021-08-05
memh5 datasets' `shape` should always be a `tuple`


anjakefala:
Two containers, the first coming directly into the class, and one being derived from it using containers.empty_like:

```
ipdb> self.mixed_data                                                                                                │
<draco.core.containers.SiderealStream object at 0x2ada05916550>                                                      │
    72.3s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector                                                │
ipdb> data                                                                                                           │
<draco.core.containers.SiderealStream object at 0x2ada0528f370>                                                      │
    78.4s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector                                                │
ipdb> data.vis                                                                                                       │
<memh5 distributed dataset '/vis': global_shape [2, 7155, 4096], dist_axis 0, type "dtype('complex64')">             │
    83.1s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector                                                │
ipdb> self.mixed_data.vis                                                                                            │
<memh5 distributed dataset '/vis': global_shape (2, 7155, 4096), dist_axis 0, type "dtype('complex64')">             │
    88.7s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector                                                │
ipdb> self.mixed_data.vis.shape                                                                                      │
(2, 7155, 4096)                                                                                                      │
    91.9s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector                                                │
ipdb> data.vis.shape                                                                                                 │
[2, 7155, 4096]                                                                                                      │
    95.8s [MPI 0/1] - DEBUG    asyncio: Using selector: EpollSelector
```

One of them returns a tuple for the shape of the vis dataset, the other returns a list, which breaks self.mixed_data.vis.shape == data.vis.shape.

PR #187, opened 2021-08-06
fix(mpiarray): ensure global_shape for new MPIArray is a tuple

anjakefala:
When following the code-threads for [this issue](https://github.com/radiocosmology/caput/issues/186), it became clear that `MPIArray` never asserts that the `global_shape` it receives is a tuple, when initially creating an `MPIArray`. This could result in [problems](https://github.com/radiocosmology/caput/issues/186), when referencing a container `dataset`, since they pull the `MemDatasetDistributed`'s `shape` from its `MPIArray's` `_global_shape`.

We may want to add `typing` information to `MPIArray`, opening PR to see if we want to do that.

At the moment, I still did not know who is changing the `tuple shape` from the original container's dataset, to the `list shape` in the `containers.empty_like` container.

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -369,7 +369,7 @@ class MPIArray(np.ndarray):
         arr = np.ndarray.__new__(cls, lshape, *args, **kwargs)
 
         # Set attributes of class
-        arr._global_shape = global_shape
+        arr._global_shape = tuple(global_shape)
         arr._axis = axis
         arr._local_shape = tuple(lshape)
         arr._local_offset = tuple(loffset)


jrs65:
> At the moment, I still did not know who is changing the `tuple shape` from the original container's dataset, to the `list shape` in the `containers.empty_like` container.

Note that it's actually the other way around. It starts as a list, and turns into a tuple.

I would guess it comes from this block (https://github.com/radiocosmology/caput/blob/2cfe06308fddcd0e5074fc596f49ff3acc9e2fe9/caput/mpiarray.py#L657) which creates a list and just initialises with that when loading from disk. 

`container.empty_like` initialises a new container via the constructor of `ContainerBase` (which calls `.add_dataset(...)`) and that correctly sends in a tuple.

anjakefala:
Ah! This explains my confusion!

jrs65 (approved):

jrs65:
I think this is the correct fix, if you think it's good to merge just go for it.

PR #188, opened 2021-08-11
feat(runner): add an option to control overwritiing old pipeline runs

jrs65:
This also changes the default setting to *disallow* overwrites.

This adds yet another feature to runner.py which is become a big mess. We probably need to do a big cleanup of it.

### DIFF
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -176,8 +176,17 @@ def run(configfile, loglevel, profile, profiler, mpi_abort, psutil):
     metavar="<NAME>=<VALUE>[,<VALUE2>...]",
 )
 @click.option("--submit/--nosubmit", default=False, help="Submit into the batch queue")
+@click.option(
+    "--overwrite",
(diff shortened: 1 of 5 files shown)


caput/scripts/runner.py
```diff
@@ -399,7 +418,21 @@ def queue(
...
+
+        if overwrite == "never":
+            click.echo(f"Job already exists at {workdir}. Skipping.")
+            return
+        elif overwrite == "failed" and statusfile.exists():
+            with open(statusfile, "r") as fh:
+                contents = fh.read()
+                if "FINISHED" in contents:
+                    click.echo(
+                        f"Successful jAob already exists at {workdir}. Skipping."
```
tristpinsm, line 432: typo

tristpinsm (approved):
As is this looks good to me. I'm not sure what the context is that prompted adding this feature, but would it also be useful to check if a successful job created an output in the destination directory?

PR #192, opened 2021-10-19
feat(mpiarray.MPIArray): Handle index selections.

tristpinsm:
closes radiocosmology/draco#137


jrs65 (changes requested):
@tristpinsm this looks good, but for this I'd like to see a test. There already exist some tests checking the ability of it to deal with slices, can you add some quick ones for it dealing with index lists?

PR #194, opened 2022-02-24
fix(memh5): ensure contained numpy arrays have a native dtype.byteorder

anjakefala:
Needed because of https://github.com/mpi4py/mpi4py/issues/177.
Shift-Bcasted arrays with little or big endian byteorders will result in
an exception.

Is it ever possible that the endian set does not match the architecture? What do we do in that case?

If we use an upgraded version of mpi4py, this PR is not needed.

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -1132,7 +1132,7 @@ class MemDatasetCommon(MemDataset):
         self = cls.__new__(cls)
         super(MemDatasetCommon, self).__init__(**kwargs)
 
-        self._data = data
+        self._data = ensure_native_byteorder(data)
         self._chunks = chunks
         self._compression = compression
         self._compression_opts = compression_opts
@@ -2526,10 +2526,17 @@ def deep_group_copy(
             # Convert strings in an HDF5 dataset into unicode
             else:
                 data = ensure_unicode(data)
+
         elif to_file:
             # If we shouldn't convert we at least need to ensure there aren't any
             # Unicode characters before writing
             data = check_unicode(entry)
+
+        if not to_file:
+            # reading from h5py can result in arrays with explicit endian set
+            # which mpi4py cannot handle when Bcasting memh5.Group
+            # needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+            data = ensure_native_byteorder(data)
 
         return data.dtype, data.shape, data
 
@@ -2581,6 +2588,7 @@ def deep_group_copy(
         else:  # Is a dataset
             dtype, shape, data = _prepare_dataset(entry)
             compression_kwargs = _prepare_compression_args(entry)
+
             g2.create_dataset(
                 key,
                 shape=shape,
@@ -2798,6 +2806,9 @@ def _distributed_group_from_file(
                         if convert_dataset_strings:
                             cdata = ensure_unicode(cdata)
 
+                        # only needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+                        cdata = ensure_native_byteorder(cdata)
+
                     cdata = comm.bcast(cdata, root=0)
 
                     # Create dataset from array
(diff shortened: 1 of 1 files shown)


jrs65 (changes requested):
The implementations of the conversion functions looks great.

However, I think all the conversions are being applied in the wrong directions. They seem to only be in the `*_to_hdf5` routines, and not in the `_from_hdf5` routines. The `_to_hdf5` routines won't be problematic as h5py will map everything to the LE types, but probably shouldn't have any conversion applied at all in case you do want to deliberately massage things.

The same is also true of `deep_group_copy`. That function has a bunch of branches such that it's a bit hard to follow, but I think it's only being applied when the destination is an `h5py.Group`, when it should actually be only applied if the conversion is `h5py.Group -> memh5.Group`.

So I think steps to do:
- [x] Remove conversions from `to_hdf5` routines.
- [ ] Add conversion to `from_hdf5` routine.
- [ ] Change `deep_group_copy` to make conversions happen only for `h5py -> memh5` copies.

anjakefala:
Attempted to address comments!

caput/memh5.py
```diff
@@ -2239,6 +2239,13 @@ def deep_group_copy(
             else:
                 data = entry[selection]
 
+            # only apply if conversion is h5py.Group -> memh5.Group
+            # reading from h5py can result in arrays with explicit endian set
+            # which mpi4py cannot handle when Bcasting memh5.Group
+            # needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+            if isinstance(g1, h5py.group) and isinstance(g2, MemGroup):
```
jrs65, line 2246: There's obviously a typo here (lower case g), which is presumably why the tests fail.

However, to prepare for a zarr world I wonder if we should change the first clause to `not instance(g1, MemGroup)`.

Also, I'm not sure what zarr does in terms of typing of arrays, so maybe we don't need to do it.

caput/memh5.py
```diff
@@ -2588,6 +2598,9 @@ def _copy_from_file(h5group, memgroup, selections=None):
 
                     cdata = comm.bcast(cdata, root=0)
 
+                    # only needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+                    cdata = ensure_native_byteorder(pdata)
```
jrs65, line 2602: This should be `cdata` as the argument, and it should also probably be before the `bcast` call. Obviously it's `Bcast` that's been causing us trouble, but just in case things change, or we change `bcast -> Bcast` at somepoint.

jrs65 (changes requested):

caput/memh5.py
jrs65: Looks like `zarr` gives an array with native order, so we don't need to do anything at all. In which case your version (with the fixed typo) is all good.

caput/memh5.py
anjakefala: Good catch on both counts! My eyes are getting tired.

anjakefala:
```
>       return self._axis
E       AttributeError: 'MPIArray' object has no attribute '_axis'
```

*grieves*

I think we should merge in the MPIArray work first, instead of coding in the workarounds.

caput/memh5.py
```diff
@@ -2569,6 +2576,9 @@ def _copy_from_file(h5group, memgroup, selections=None):
                         h5group, key, axis=distributed_axis, comm=comm, sel=selection
                     )
 
+                    # only needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+                    pdata = ensure_native_byteorder(pdata)
```
jrs65, line 2580: I'm pretty sure this call is the one that's causing the test failures, but it actually isn't needed as `.from_hdf5` will guarantee that pdata has a explicit native endianness. Maybe true removing it, and see if things work in that case?

jrs65 (changes requested):

caput/memh5.py
```diff
@@ -2586,6 +2593,9 @@ def _copy_from_file(h5group, memgroup, selections=None):
                         if convert_dataset_strings:
                             cdata = ensure_unicode(cdata)
 
+                    # only needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
+                    cdata = ensure_native_byteorder(cdata)
```
anjakefala, line 2597: Only rank 0 will have `cdata`.

jrs65 (approved):

PR #197, opened 2022-06-06
fix(mpiarray): handling Ellipsis and np.newaxis when slicing

jrs65:


jrs65:
Just fixed up a few documentation issue I found, and restructured the commits

PR #198, opened 2022-06-15
HFB container fixes

jrs65:


caput/memh5.py
```diff
@@ -2514,7 +2514,7 @@ def _prepare_dataset(dset):
                 return dset.dtype, dset.shape, None
 
         # Extract the data for the selection
-        data = entry[selection]
```
tristpinsm, line 2518: huh, so I guess this just worked because it was being called in such a way that `entry` was correctly defined. How did you spot this?

tristpinsm (approved):

caput/memh5.py
jrs65: Just by reading through the code nearby. But yeah, you're write that it's only the fact the variable if function scoped that it worked.

caput/memh5.py
```diff
@@ -1770,7 +1770,7 @@ def from_file(
                 data = file_
                 toclose = False
             else:
-                kwargs.setdefault("mode", "a")
+                kwargs.setdefault("mode", "r")
```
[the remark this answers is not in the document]
jrs65: I don't think so. I think it's not that often that we use this functionality, so it's probably fairly risk free.

PR #200, opened 2022-06-17
fix(mpiarray): crash testing for an Ellipsis when given an ndarray index

jrs65:


tristpinsm (approved):
These changes make sense, but as far as I can tell the new test and the changes to the ellipsis handling are not related. Should there be two commits?

jrs65:
Yeah, they are related but it's a bit subtle. The problem with the old code was that if you pass in a numpy array as an array index it does an element wise comparison to Ellipsis and then complains that you need to use `.all()` or `.any()` on the result. So this test now checks that case and thus throws an error with the previous implementation.

ljgray:
Any reason this hasn't been merged? Ran into the Ellipsis crash issue so having to use this branch to get around it.

PR #202, opened 2022-06-30
fix(mpiarray): only give warning when compression is disabled by the pipeline

jrs65:
This PR cleans up the logic around how the writing code decides whether to use collective IO or not, and then only warns for unexpected cases.


tristpinsm (approved):
The logic is also a lot easier to understand now

PR #204, opened 2022-07-22
MPIArray cleanups and speedups

jrs65:
- test(mpiarray): tests actually cover __call__ ufuncs not outer ones
- test(mpiarray): check reductions over multiple axes
- test(mpiarray): increase test coverage
- fix(mpiarray): address ufunc corner cases and remove unncessary MPI calls

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -414,16 +414,18 @@ class MPIArray(np.ndarray):
 
             if isinstance(dist_axis_index, int):
                 warnings.warn(
-                    "You are indexing directly into the distributed axis."
-                    "Returning a view into the local array."
-                    "Please use global_slice, or .local_array before indexing instead."
+                    "You are indexing directly into the distributed axis "
+                    "returning a view into the local array. "
+                    "Please use global_slice, or .local_array before indexing instead.",
+                    stacklevel=2,
                 )
 
                 return self.local_array.__getitem__(v)
             warnings.warn(
-                "You are directly sub-slicing the distributed axis."
-                "Returning a view into the local array."
-                "Please use global_slice, or .local_array before indexing."
+                "You are directly sub-slicing the distributed axis "
+                "returning a view into the local array. "
+                "Please use global_slice, or .local_array before indexing.",
+                stacklevel=2,
             )
             return self.local_array.__getitem__(v)
 
@@ -1791,7 +1793,8 @@ class MPIArray(np.ndarray):
                         "A ufunc is combining an MPIArray with a numpy array that "
                         "matches exactly the local part of the distributed axis. This "
                         "is very fragile and may fail on other ranks. "
-                        f"Numpy array shape: {inp.shape}; dist axis: {cur_dist_axis}."
+                        f"Numpy array shape: {inp.shape}; dist axis: {cur_dist_axis}.",
+                        stacklevel=3,
                     )
                 elif cur_axis_length != 1:
                     raise AxisException(
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1380,6 +1380,22 @@ class MPIArray(np.ndarray):
 
         return rdata
 
+    @property
+    def _native_type(self) -> bool:
+        """Is this array a type understood by mpi4py."""
+        try:
+            _ = mpiutil.typemap(self.dtype)
+            return True
+        except KeyError:
+            return False
+
+    def _prep_buf(self, x: np.ndarray) -> np.ndarray:
+        """Prepare a buffer for sending/recv by turning into a view of the bytes."""
+        if self._native_type:
+            return x
+        else:
+            return x.view(np.byte).reshape(x.shape + (self.itemsize,))
+
     def gather(self, rank=0):
         """Gather a full copy onto a specific rank.
 
@@ -1393,6 +1409,7 @@ class MPIArray(np.ndarray):
         arr : np.ndarray, or None
             The full global array on the specified rank.
         """
+
         if self.comm.rank == rank:
             arr = np.ndarray(self.global_shape, dtype=self.dtype)
         else:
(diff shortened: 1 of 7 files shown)


caput/mpiarray.py
```diff
@@ -1602,98 +1610,201 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
 
-        if ufunc.nout == 1:
-            results = (results,)
+            # Check the distribution of the output arguments makes sense
+            for ii, array in enumerate(out_args):
+                try:
+                    _check_dist_axis(array, new_dist_axis, global_length, offset, comm)
+                except Exception as e:
+                    raise ValueError(
+                        "Output argument at position {ii} does not match expectation."
```
ljgray, line 1699:
```suggestion
                        f"Output argument at position {ii} does not match expectation."
```

jrs65: Good spot. Thanks!

caput/mpiarray.py
```diff
@@ -1602,98 +1609,203 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
+
+        # For a 'call' the highest dimensional array should determine the dimensionality
+        # of the output, so we find that
+        max_dim_input = inputs[
+            np.argmax(
+                [inp.ndim if isinstance(inp, np.ndarray) else -1 for inp in inputs]
+            )
+        ]
+
+        if not isinstance(max_dim_input, MPIArray):
```
ljgray, line 1740: This appear to cause issues in unintended places. Tested in test_daily pipeline (python 3.10.2, numpy 1.22.2) results in the following error:
```
   139.7s [MPI 0/4] - DEBUG    caput.pipeline: Task ComputeSystemSensitivity calling 'next()'.
   139.7s [MPI 0/4] - INFO     draco.analysis.sensitivity.ComputeSystemSensitivity: Starting next for task ComputeSystemSensitivity
...
  File "/project/6003614/lgray/chime_pipeline_dev/code/draco/draco/analysis/sensitivity.py", line 213, in process
    radiometer[:, pp, :] += nsq * auto[:, ii, :] * auto[:, jj, :]
  File "/project/6003614/lgray/chime_pipeline_dev/code/caput/caput/mpiarray.py", line 1686, in __array_ufunc__
    new_dist_axis, global_length, offset = _validation_fn(ufunc, *inputs, **kwargs)
  File "/project/6003614/lgray/chime_pipeline_dev/code/caput/caput/mpiarray.py", line 1741, in _array_ufunc_call
    raise ValueError("Highest dimensional input must be an MPIArray.")
ValueError: Highest dimensional input must be an MPIArray.
```

ljgray:
I'm wondering if there should be an automatic fallback to the old method of calling ufuncs in the event that this fails, at least for now. Adding additional restrictions to the implementation might cause other unintended errors like the one I saw above.

caput/mpiarray.py
jrs65: Thanks! I hadn't yet gotten around to running `test_daily` on it yet. I'll fix that issue and try it again myself.

caput/mpiarray.py
```diff
@@ -1602,97 +1610,226 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
         ret = []
+        for res, out in zip(results, out_args):
+            # If out is None a new *numpy* array was initialised for the results. We
+            # need to turn this into an MPIArray using the inferrred distribution
+            if out is None:
+                # Result is a scalar, so we want to return this as a 1D distributed
+                # vector
+                if not res.shape:
+                    res = res.reshape(1)
+
```
ljgray, line 1720: Just confirming that all values returned from super().__array_ufunc__ are numpy objects, otherwise this could fail (if something doesn't have a .reshape routine)

ljgray (approved):
Seems good from a look through. It might be worth doing a dive into mpiarray as a whole in a future revision, just to make sure that it doesn't blow up and become too difficult to maintain.

ljgray:
I'm curious to see profiling on this since it adds some extra python overhead. Probably won't be too significant.

jrs65:
> I'm wondering if there should be an automatic fallback to the old method of calling ufuncs in the event that this fails, at least for now. Adding additional restrictions to the implementation might cause other unintended errors like the one I saw above.

I vaguely thought about that, but in the end I think probably not. Mostly because the old code turned out to have a large number of pathways to generating corrupt or non-sensical MPIArrays, and doing operations which were pretty ill defined how they should work.

It's definitely true that this new implementation is a bit more strict, and I've found one place that I needed to fix up the pipeline code to make it work, but that's probably a good thing.

jrs65:
> I'm curious to see profiling on this since it adds some extra python overhead. Probably won't be too significant.

I'm intrigued. I think there probably isn't much extra overhead, as it actually doesn't do much more in total for validation, it just branches out on the method and does different validation for each.

caput/mpiarray.py
jrs65: Good point. However, I think as you suggest it is guaranteed that even scalar values are numpy objects and have both `.shape` and `.reshape()`.

jrs65:
I'm going to delay merging for the moment, just in case I need to push any more fixes.

ljgray:
> > I'm curious to see profiling on this since it adds some extra python overhead. Probably won't be too significant.
> 
> I'm intrigued. I think there probably isn't much extra overhead, as it actually doesn't do much more in total for validation, it just branches out on the method and does different validation for each.

True, I was thinking about it compared to standard numpy ufuncs, which would have almost no overhead, but this new implementation doesn't really look like it would be much different from the previous one.

jrs65:
Yeah, it would be really interesting to see that profiling. It probably depends exactly which ufunc you use, but I'm interested as to how big an array you need for the time cost of the actual numerics to dominate over the overhead.

ljgray (approved):
I think I'm mostly up to date on what has been changed here, I'll do a deeper dive over the weekend to see  if there are any minor issues, but I ran a test pipeline last night without issue, so I think things are generally working well.

Issue #205, opened 2022-07-28
`numpy.median` fails on mpi arrays


ljgray:
A TypeError is thrown when calling `numpy.median` on an MPIArray. 

The root issue is that `numpy.median` tries to pass "order=K" as an argument to `asanyarray(x).copy()`, but `MPIArray.copy` does not accept any arguments. the `copy` method is also called when taking a mean, but no argument is passed so no errors occur. Passing the argument through to the underlying `numpy.copy` call results in "IndexError: list assignment index out of range"

Edit: the above issue is not actually the underlying problem. `numpy.median` fails on mpi arrays due to some underlying mechanism

PR #206, opened 2022-07-28
fix `np.median` error with MPIArray

ljgray:
…ssed through to the underlying numpy.copy call

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1465,7 +1465,25 @@ class MPIArray(np.ndarray):
 
         return arr
 
+    def copy(self, *args, **kwargs):
+        """Return a copy of the MPIArray.
+
+        Returns
+        -------
+        arr_copy : MPIArray
+        """
+        return MPIArray.wrap(
+            self.view(np.ndarray).copy(*args, **kwargs), axis=self.axis, comm=self.comm
+        )
+
     def ravel(self, *args, **kwargs):
+        """This method is explicitly not implemented and will
+        return a NotImplementedError.
+
+        Raises
+        ------
+        NotImplementedError
+        """
         raise NotImplementedError(
             "Method 'ravel' is not implemented for distributed arrays. "
             "Try using '.local_array'."
@@ -1873,8 +1891,6 @@ class MPIArray(np.ndarray):
             raise AxisException(
                 f"Distributed axis {axis} does not exist in global shape {global_shape}"
             ) from e
-
-        print(global_shape, axis, axlen)
 
         global_shape[axis] = axlen
 
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1378,25 +1378,6 @@ class MPIArray(np.ndarray):
 
         return rdata
 
-    # def copy(self, *args, **kwargs):
-    #     """Return a copy of the MPIArray.
-
-    #     Returns
-    #     -------
-    #     arr_copy : MPIArray
-    #     """
-    #     # global_len = len(self.view(np.ndarray)[self.axis])
-    #     # local_start = 0
-
-    #     return MPIArray.wrap(
-    #     # return MPIArray._view_from_data_and_params(
-    #         self.view(np.ndarray).copy(*args, **kwargs), axis=self.axis, 
-    #         # global_length=global_len,
-    #         # local_start=local_start,
-    #         comm=self.comm,
-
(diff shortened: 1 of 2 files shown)


ljgray:
This should fix #205

caput/mpiarray.py
```diff
@@ -1378,15 +1378,15 @@ def reshape(self, *shape):
...
+    def copy(self, *args, **kwargs):
         """Return a copy of the MPIArray.
 
         Returns
         -------
         arr_copy : MPIArray
         """
         return MPIArray.wrap(
-            self.view(np.ndarray).copy(), axis=self.axis, comm=self.comm
+            self.view(np.ndarray).copy(*args, **kwargs), axis=self.axis, comm=self.comm
```
jrs65, line 1389: As you can guarantee this is properly formed you can probably replace the wrap with `_view_from_data_and_params` which would save some collective communication.


Finally, it *might* be possible to remove this copy method entirely and just reply on using numpys directly. Now we've correctly implemented all the area creation hooks numpys own copy method might just work as-is.

jrs65 (changes requested):

caput/mpiarray.py
ljgray: I'll give both a try

ljgray:
So unfortunately neither of those options fix the issue. The underlying problem seems to be in `__array_finalize__`. Calling `np.median` on an MPIArray calls `__array_finalize__` 3 times. On the second call, the dimension of local_shape is reduced by 1, and on the final call the dimension is reduced to 1. So, if the distributed axis is anything other than 0, the final call will fail because it tries to index into `global_shape`, which has dimension 1, on a higher axis.

jrs65:
I'm confused. I don't why the copy could work using the wrap, but not via `_view_from_data_and_params`. How are you using it? I would have guessed something like this would do it:
```python
return MPIArray. _view_from_data_and_params(
    self.local_array.copy(*args, **kwargs), self.axis, self.global_shape[self.axis], self.local_offset[self.axis], self.comm
)
```
That's essentially what a wrap call does internally, but it skips the checks as you know you don't need them.

ljgray:
> I'm confused. I don't why the copy could work using the wrap, but not via `_view_from_data_and_params`. How are you using it? I would have guessed something like this would do it:
> 
> ```python
> return MPIArray. _view_from_data_and_params(
>     self.local_array.copy(*args, **kwargs), self.axis, self.global_shape[self.axis], self.local_offset[self.axis], self.comm
> )
> ```
> 
> That's essentially what a wrap call does internally, but it skips the checks as you know you don't need them.

Sorry, it doesn't seem to work either way, whether using wrap, _view..., or the numpy copy method. But I don't think that this is actually the issue, as mentioned in my comment above.

ljgray:
I believe that the underlying issue here is a `.ravel()` call in `numpy._median_nancheck`, which doesn't really work for a distributed array.

ljgray:
I think that this should work properly now. The .ravel() call throws a NotImplementedError.

caput/mpiarray.py
```diff
@@ -1476,6 +1465,22 @@ def allgather(self):
...
+        """Return a copy of the MPIArray.
+        Returns
+        -------
+        arr_copy : MPIArray
+        """
+        return MPIArray.wrap(
+            self.view(np.ndarray).copy(*args, **kwargs), axis=self.axis, comm=self.comm
+        )
+
+    def ravel(self, *args, **kwargs):
```
jrs65, line 1478: Can you add a docstring here? Just so that if someone only reads the documentation they'll be able to tell that `.ravel` doesn't work, rather than seeing it's presence and inferring that it doesn't work.

jrs65 (changes requested):
Looks good to me, I think just the docstring comment.

Also, can you rebase all the commits down to just the externally relevant changes before merging?

caput/mpiarray.py
ljgray: Oh, good catch.

caput/mpiarray.py
```diff
@@ -1476,6 +1465,22 @@ def allgather(self):
 
         return arr
 
+    def copy(self, *args, **kwargs):
```
jrs65, line 1468: Also there's a formatting issue in this docstring.

jrs65 (changes requested):

PR #207, opened 2022-08-02
Add explicit TypeError when ufunc tries to output to a numpy array

ljgray:
Although this should be fixed everywhere in the pipeline at the moment, it is not very clear that the actual error that ends up getting thrown is caused by this specifically. This check might help us in the future by making it clear what is going on.

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1703,6 +1703,12 @@ class MPIArray(np.ndarray):
         # arrays
         if "out" in kwargs:
             out_args = kwargs.get("out", None)
+            if not all(isinstance(x, MPIArray) for x in out_args):
+                raise TypeError(
+                    "At least one output is not an MPIArray. This can happen if a ufunc "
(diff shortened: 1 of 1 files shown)


caput/mpiarray.py
```diff
@@ -1703,6 +1703,12 @@ def __array_ufunc__(
         # arrays
         if "out" in kwargs:
             out_args = kwargs.get("out", None)
+            if not all(isinstance(x, MPIArray) for x in out_args):
+                raise TypeError(
+                    "At least one output is not an MPIArray. This can happen if a ufunc "
+                    "is trying to modify a np.array in-place using values from a MPIArray. "
+                    "Try using .local_array or cast the np.array to MPIArray."
```
jrs65, line 1710:
```suggestion
                    "is trying to modify a np.ndarray in-place using values from a MPIArray. "
                    "Try using .local_array or cast the np.ndarray to MPIArray."
```

jrs65 (approved):
Looks good to me. Thanks!

PR #210, opened 2022-08-17
feat(MPIArray): methods for left and right padding from adjacent rank

ljgray:
'left_pad' and 'right_pad' methods return n rows along the distributed
axis from left and right adjacent ranks, respectively. The returned
arrays are filled with NaNs to guarantee shape.

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -575,6 +575,82 @@ class MPIArray(np.ndarray):
         """
         return self.view(np.ndarray)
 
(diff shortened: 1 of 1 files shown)


ljgray:
I'm still experimenting a bit here but input is valuable, particularly if there is a better way to accomplish this that is being missed here.

jrs65:
My comment is that this seems like a complex scheme. Why not just allgather the full array (which isn't very large), and then limit the range of the rolling median computation?

PR #211, opened 2022-08-17
feat(MPIArray): property for getting global bounds of local array.

ljgray:


caput/mpiarray.py
```diff
@@ -575,6 +575,22 @@ def local_array(self):
...
+    @property
+    def local_bounds(self):
+        """
+        Global bounds of the local array along the
+        distributed axis.
+
+        Returns
+        -------
+        local_bounds : slice
+        """
```
jrs65, line 587:
```suggestion
    def local_bounds(self) -> slice:
        """
        Global bounds of the local array along the
        distributed axis.

        Returns
        -------
        local_bounds
        """
```

jrs65 (approved):
Great. I'm stoked! We should have done this years ago. One small suggested change to the doc/annotation, but you should decide what you want to do with it and then feel free to merge without another review.

PR #212, opened 2022-08-23
build(setup.py, .github/workflows/main.yml): un-pin numcodecs version

sjforeman:
`numcodecs 0.7.3` uses a version of `py-cpuinfo` that does not support Apple Silicon. Newer versions of `numcodecs` don't have this issue, so if possible, we should not pin to 0.7.3.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,7 +62,7 @@ jobs:
         pip install h5py
         pip install -r requirements.txt
         pip install zarr==2.8.1
-        pip install mpi4py numcodecs==0.7.3 bitshuffle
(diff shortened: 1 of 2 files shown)


sjforeman:
The CI checks all give the same results as with `numcodecs==0.7.3`, so presumably this change is benign, but let me know if there are explicit checks I should carry out

jrs65 (approved):
Did you find any indication of why we pinned it in the first place?

sjforeman:
I couldn't find a good reason to require `numcodecs==0.7.3`, no: the changes since that version (https://numcodecs.readthedocs.io/en/stable/release.html) don't look very drastic. I just updated this PR to require at least 0.7.3, though, in case there might be untested issues with older versions.

PR #219, opened 2022-09-22
docs(pipeline): add description of outputs of python-based pipeline

sjforeman:


jrs65 (approved):
Thanks! My favourite thing in the world is when someone takes it upon themselves to improve the documentation.

PR #221, opened 2022-09-29
Move cython `_invert_no_zero` into `caput`

ljgray:
Multiple versions of `_invert_no_zero` existed, with a faster cython version existing in `draco`. Since `draco` is not a dependency everywhere, this PR would move `_invert_no_zero` into `caput`. 

Accompanying PRs:
radiocosmology/draco#202
chime-experiment/ch_util#41

### DIFF
--- a/caput/_fast_tools.pyx
+++ b/caput/_fast_tools.pyx
@@ -16,7 +16,7 @@ ctypedef fused real_or_complex:
 @cython.wraparound(False)
 @cython.boundscheck(False)
 @cython.cdivision(True)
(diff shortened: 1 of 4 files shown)


ljgray:
Basically just directly copying over the method from `draco`. Note that radiocosmology/draco#202 and chime-experiment/ch_util#41 will continue to fail tests until this is merged.

jrs65 (changes requested):

jrs65 (approved):

PR #225, opened 2022-10-13
feat(fileformats): make guess_file_formats compatible with pathlib.Pa…

ljgray:
…th objects


jrs65 (approved):
Yay. I kept meaning to fix this one.

PR #227, opened 2022-11-11
Observer property to get current LSD/CSD

ljgray:
Also add a helper method for getting unique values from a list while maintaining order. This should be pretty fast without relying on numpy

### DIFF
--- a/caput/time.py
+++ b/caput/time.py
@@ -228,6 +228,10 @@ class Observer:
 
     _obs = None
 
+    def get_current_lsd(self) -> float:
+        """Get the current LSD."""
+        return self.lsd(datetime_to_unix(datetime.utcnow())).astype(np.float64)
+
     def skyfield_obs(self):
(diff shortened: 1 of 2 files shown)


caput/time.py
```diff
@@ -228,6 +228,11 @@ def __init__(self, lon=0.0, lat=0.0, alt=0.0, lsd_start=None, sf_wrapper=None):
 
     _obs = None
 
+    @property
```
jrs65, line 231: The implementation looks good, but I'm not sure if this should be a property for reasons related to the comments in the other PR.

jrs65 (approved):
I've approved, contingent on you figuring out what you want to do with this.

caput/time.py
ljgray: Given that there isn't any specific reason why it needs to be a property, I'm going to remove that

PR #229, opened 2022-11-18
Add macos-latest to ci pipeline

ljgray:
Depends on https://github.com/kiyo-masui/bitshuffle/pull/155


ljgray:
Right I guess this should

ljgray:
I'm not familiar with using brew on macos but I think it will have to replace the apt-get step

jrs65 (approved):

jrs65 (changes requested):
Actually this needs to actually run properly first. So let me suggest some changes.

.github/workflows/main.yml
```diff
@@ -41,16 +41,27 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
 
-    - name: Install apt dependencies
+    - name: Install apt/brew dependencies
       run: |
-        sudo apt-get update
-        sudo apt-get install -y libopenmpi-dev openmpi-bin libhdf5-serial-dev
+        ${{ matrix.os-installer }} ${{ matrix.os-update }}
+        ${{ matrix.os-installer }} ${{ matrix.os-dependencies }}
```
jrs65, line 64: I think this would be easier to follow and customise if rather than having a single install step with the matrix customisations you just had two steps (Install apt; Install brew) which ran conditionally on each OS.

On macOS I wouldn't bother with the update-reset step which you would be able to skip if you broke it up

jrs65 (dismissed):

ljgray:
Yes haha I've been having issues getting Brew to install/set up some dependencies properly so I've just kept this as a draft

ketiltrout (approved):
This is great.

ketiltrout:
I've updated the list of required checks in the branch protection rules

Issue #190, opened 2021-10-17
`memh5.from_group` does not have the expected behavior


leungcalvin:
This bug was discovered downstream (in `baseband_analysis`) whose main data container `BBData` inherits from `BasicCont`. I tried to make a new, independent instance of a `BBData` object that lives in memory. I tried doing this using `from_group` but it looks like the new object created is, in fact, not a deep copy of the original (even though `from_group` calls something which is supposed to be `deep_group_copy`).

Modifying the datasets of the new object modifies the original. Is this the expected behavior? This is potentially dangerous for downstream analysis; tagging @jrs65 and @kiyo-masui because I'm so out of my depth here.

![image](https://user-images.githubusercontent.com/42367504/137636700-bdbb21b5-311d-4c60-80bf-0d42df84e0b3.png)

jrs65:
For CHIME cosmology we built this functionality in at a higher level of abstraction into [draco.core.containers.ContainerBase.copy](https://github.com/radiocosmology/draco/blob/5d3bd9144d37f6cac9c80d4f07c242c699e177b0/draco/core/containers.py#L408), so we've never really noticed it was missing.

I think we could feasibly move the functionality down to a lower level (probably BasicCont) though if that would be helpful. It would require a little bit of refactoring, but provided the API remained the same I don't think we would care.

ljgray:
@ketiltrout

PR #236, opened 2023-03-01
Provide `.copy` methods for `MemDataset` and `MemDiskGroup`

ljgray:
This PR provides `.copy` methods for `MemDataset` and `MemDiskGroup` classes. 

`MemDataset.copy` can optionally do a deep or shallow copy (deep is default) and set the memory layout of the underlying data (if it is a numpy array). By default, this is the same layout as the dataset being copied. 

`MemDiskGroup.copy` deep copies all datasets by default and provides a `shared` argument. This is implemented via the `deep_group_copy` method, which is extended to optionally deep copy datasets (except those that are shared) and copy distributed arrays __only__ in the memory -> memory case. I made considerations based on discussion in #190. 

This PR also provides a a tool to check two objects which may contain numpy arrays for equality (ex: a list of arrays, dict with array arguments, etc...). For example, deep copying a list which contains a numpy array would result in an error when comparing the dicts, as the numpy arrays would not be the same object and would fall back on elementwise comparison. Since `list.__eq__` won't call `all()` on the numpy array after elementwise comparison, the comparison would fail. The same thing happens if the outer container is a dict.

Finally, the PR adds a check for matching `dtype` when comparing two `MemDatasets`. I'm undecided if this is a good idea or not.

NOTE: I've done a fair bit of testing and I think this is tentatively ready for review. I do still want to try to break it a bit more thoroughly. I've tagged a few possibly relevant people for review but it's fine if this waits for awhile

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -62,6 +62,7 @@ import logging
 
 import numpy as np
 import h5py
+from copy import deepcopy
 
 from . import fileformats
 from . import mpiutil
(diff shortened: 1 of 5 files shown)


caput/memh5.py
```diff
@@ -2534,7 +2603,14 @@ def _prepare_dataset(dset):
             # needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
             data = ensure_native_byteorder(data)
 
-        return data.dtype, data.shape, data
+        dset_args = {"dtype": data.dtype, "shape": data.shape, "data": data}
+        # If we're copying memory to memory we can allow distributed datasets
+        if not to_file and isinstance(dset, MemDatasetDistributed):
+            dset_args.update(
```
ljgray, line 2609: Need to be careful with this as it could change some behaviour

caput/memh5.py
```diff
@@ -2601,6 +2688,38 @@ def _prepare_compression_args(dset):
     return distributed_dset_names if skip_distributed else None
```
ljgray, line 2690: I considered using `@overload` here, but it really doesn't look pretty. I'm open to thoughts about using something better than `Any` though

caput/memh5.py
```diff
@@ -2822,6 +2941,9 @@ def _copy_from_file(h5group, memgroup, selections=None):
     return group
 
 
+# Some extra functions for types. Should maybe move elsewhere
+
```
ljgray, line 2945: I'm debating moving everything below to a new file since their utility is grouped together quite well and it adds a lot of clutter to `memh5`. Still undecided though

kiyo-masui (commented):
I've given this a full read, and it looks good to me. I'll leave final approval to @jrs65 since he is much more up-to-date on this code than I am at this point.

caput/tools.py
```diff
@@ -66,3 +66,71 @@ def unique_ordered(x: Iterable) -> list:
...
+            _assert_equal(obj1[k], obj2[k])
+
+        return
+
+    # Ensure that any np.ndarray subclass is an np.ndarray
+    if isinstance(obj1, np.ndarray):
+        obj1 = obj1.view(np.ndarray)
+
+    if isinstance(obj2, np.ndarray):
+        obj2 = obj2.view(np.ndarray)
```
jrs65, line 133: What do you want to happen if someone does this?
```python
arr1 = mpiarray.MPIArray((4,) axis=0, comm=MPI.COMM_SELF)
arr1.local_array[:] = np.arange(4)

arr2 = np.arange(4)

tools._assert_equal(arr1, arr2)
```
It seems like you want them to evaluate to equal, but I'm not sure if that should generally be considered to be true.

jrs65: Ah I see you even wrote a test to cover this one. I'm still not sure if this is generally desirable though.

ljgray: I see the issue with having that compare to be equal. I'll tweak it a bit. 

Annoyingly, the only reason we have to take the view in the first place is because `numpy.testing.assert_equal` calls `.ravel` on `ndarray` subclasses, which obviously breaks with a `MPIArray`.

PR #237, opened 2023-03-10
Allow fileformat to be set by fileformat subclass

ljgray:
Setting a task `output_format` config property fails if given anything except `h5py` or `zarr`. This change allows the `output_format` config to be set by a `fileformats.FileFormat` subclass (`H5PY` or `Zarr`).

I should maybe take a deeper dive into this, because it actually shouldn't have been hitting the `val = val.lower()` call anyway...

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -51,6 +51,11 @@ Richard 40.0
 Richard 40.0 Sooty
 
 """
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from . import fileformats
 
 import logging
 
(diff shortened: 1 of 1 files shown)


caput/config.py
```diff
@@ -514,16 +514,21 @@ def _prop(val):
         if val is None:
             return None
 
+        if issubclass(val, fileformats.FileFormat):
+            return val
+
         if not isinstance(val, str):
             CaputConfigError(
```
sjforeman, line 521: While we're editing this file, we should probably fix:
```suggestion
            raise CaputConfigError(
```

sjforeman (approved):
Looks good to me. Perhaps we should also edit the docstring of `file_format` to reflect the added functionality?

caput/config.py
ljgray: Good catch! I was trying to figure out why it failed at the call to `val.lower()` and not here, and I somehow completely missed that this wasn't actually being raised!

PR #239, opened 2023-03-20
fix(memh5): fix a bug where shared datasets weren't actually shared

ljgray:
This PR makes the `shared` argument in `MemDiskGroup` actually share the `MemDataset` object and not just the underlying data array. Based on the idea of shared datasets being maintained through operations such as redistribution, axis downselection is not allowed for shared datasets.

It's important to note that the __default__ behaviour of `memh5.deep_group_copy` is (and has been since inception) to only make a _shallow_ copy of the `MemDataset` - i.e., the underlying array is actually shared. It doesn't matter when using this method to copy from disk -> memory or memory-> disk, but it does have implications when copying memory -> memory. This _probably_ shouldn't be the default behaviour, but we would have to be extremely careful if making any changes to avoid breaking anything.  

I've also updated the docstring for `deep_group_copy` to make note of the copy behaviour.

This lets us remove the `.copy` method in `draco.ContainerBase` as it will properly replicate that behaviour (see radiocosmology/draco#231)

~~Also, add an option to skip specific datasets when copying.~~

### DIFF
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2701,8 +2701,20 @@ def deep_group_copy(
     # If copying to file, datasets are not shared, so ensure that these
     # datasets are properly processed
     if to_file:
-        shared = {}
-        shallow = False
+        if shared:
+            warnings.warn(
+                f"Attempted to share datasets {(*shared,)}, but target group "
+                f"{g2} is on disk. Datasets cannot be shared."
+            )
+            shared = {}
+
+        if shallow:
+            warnings.warn(
+                f"Attempted to make a shallow copy of group {g1}, but target "
+                f"group {g2} is on disk. Datasets cannot be shared."
+            )
+            shallow = False
+
     elif not shallow:
         # Make sure shared dataset names are properly formatted
         shared = {"/" + k if k[0] != "/" else k for k in shared}
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2655,12 +2655,6 @@ def deep_group_copy(
             # Unicode characters before writing
             data = check_unicode(entry)
 
-        if not to_file:
-            # reading from h5py can result in arrays with explicit endian set
-            # which mpi4py cannot handle when Bcasting memh5.Group
-            # needed until fixed: https://github.com/mpi4py/mpi4py/issues/177
-            data = ensure_native_byteorder(data)
-
         dset_args = {"dtype": data.dtype, "shape": data.shape, "data": data}
         # If we're copying memory to memory we can allow distributed datasets
         if not to_file and isinstance(dset, MemDatasetDistributed):
--- a/caput/memh5.py
+++ b/caput/memh5.py
@@ -2051,9 +2051,7 @@ class MemDiskGroup(_BaseGroup):
         """
         cls = self.__class__.__new__(self.__class__)
         MemDiskGroup.__init__(cls, distributed=self.distributed, comm=self.comm)
-        deep_group_copy(
-            self._data, cls._data, deep_copy_dsets=not shallow, shared=shared
-        )
+        deep_group_copy(self._data, cls._data, shallow=shallow, shared=shared)
 
         return cls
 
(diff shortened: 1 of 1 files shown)


jrs65:
I have just been looking through this one again and reflecting on what the behaviour should be. It's not clear to me at what level a shared dataset should be shared. Really the only purpose of it is to save on memory, so anything that does that is fine. I'm not sure about shared datasets maintaining that through redistribution though, it also might be good to allow shared selections but that has its own implications.

ljgray:
I think that it's valid for a user to expect that a shared dataset would remain shared through a redistribution, since that's a common operation and it defeats the purpose of memory savings if we lose those savings the moment we redistribute. 

Selection is tricky since it would effectively have to replace the memdataset within the group with a new one with the selections applied, so I'm not sure how that would work without modifying the way memdatasets are referenced by the group (i.e., rather than the shared dataset just pointing to the same object, it would somehow have to point to a reference in the group tree which in turn references a dataset object, so if the dataset object changed the tree reference would still be the same thing). Unless you just mean applying selections during the copy, which has the potential to be tricky as well and probably wouldn't be used enough to be worth the effort (right now someone could just make the selection then make a shared copy, which would work fine)

caput/memh5.py
```diff
@@ -2686,15 +2723,30 @@ def _prepare_compression_args(dset):
...
                 g2.create_group(key)
+
             stack += [entry[k] for k in sorted(entry, reverse=True)]
-        else:  # Is a dataset
+
+        elif key in skip:
+            # Ignore this dataset
+            continue
+
+        elif key in shared:
```
ketiltrout, line 2733: I haven't looked at this in depth yet, but I'm worried here that introducing a separate code path for datasets in `shared` means that the last part of the description for the `shared` argument is no longer true: 

>    shared : list, optional
>        List of datasets to share, if `deep_copy_dsets` is True. Otherwise, no effect.

i.e. passing in a `shared` list does seem to have an effect even if `deep_copy_dsets` isn't True.  It seems like, for datasets, in `shared`, there's a lot of stuff in `_prepare_dataset` that isn't happening anymore (like the `to_file` check).  Also `convert_attribute_strings` seems to be ignored for these datasets.

Maybe none of these things are relevant to shared datasets, but if that's the case the documentation needs to better reflect what's actually going on.

ljgray:
@ketiltrout I've updated this to simplify the copy pathways and address the `to_file` issues you mentioned. Now, there should only be two paths:
- If shallow is False (default), all datasets are deep-copied *except* those in `shared`. In the case where `to_file` is True, dataset names provided in `shared` are ignored
- If shallow is True, all datasets are fully shared - no processing, data type changes, etc... are applied. We end up with a new group pointing to the same `MemDataset` objects as the original group.

ljgray:
@ketiltrout I also removed an extra byte order check, which was required for a now-fixed mpi4py bug.

caput/memh5.py
```diff
@@ -2558,14 +2564,16 @@ def deep_group_copy(
...
-    shared : list, optional
-        List of datasets to share, if `deep_copy_dsets` is True. Otherwise, no effect.
+    shallow : bool, optional
+        Explicitly share all datasets. This will only alter behaviour when copying
+        from memory to memory. If False, any dataset listed in `shared` will NOT be copied.
+        Default is False.
+    shared : iterable, optional
+        Iterable (list, set, generator) of datasets to share, if `deep_copy_dsets` is True.
+        Otherwise, no effect. Shared datasets just point to the existing object in g1 storage.
+        Axis selections cannot be applied to shared datasets.
```
ketiltrout, line 2576:
```suggestion
    shared : iterable, optional
        Iterable (list, set, generator) of datasets to share, if `shallow` is False. Shared
        datasets just point to the existing object in g1 storage.  Axis selections cannot
        be applied to shared datasets.    Ignored if `shallow` is True, since, in that case,
        _all_ datasets are shared.
```

ketiltrout (approved):
Yeah, this looks a lot better!  I've made a few suggestions for your docstring.

One potential thing I've though of:

In the case when `g2` is a file, instead of ignoring `shared`, another option would be to raise a `ValueError` on a non-empty value.  I don't know if that's better or worse than what you've implemented.

`ValueError` could help users find catch where they've maybe misconstrued their call to `deep_group_copy`, but the downside is it forces any call to `deep_group_copy` to be type-aware (i.e. of the type `g2`), which may be a worse problem.

ljgray:
> Yeah, this looks a lot better! I've made a few suggestions for your docstring.
> 
> One potential thing I've though of:
> 
> In the case when `g2` is a file, instead of ignoring `shared`, another option would be to raise a `ValueError` on a non-empty value. I don't know if that's better or worse than what you've implemented.
> 
> `ValueError` could help users find catch where they've maybe misconstrued their call to `deep_group_copy`, but the downside is it forces any call to `deep_group_copy` to be type-aware (i.e. of the type `g2`), which may be a worse problem.

I've decided to meet in the middle and add a warning. I don't really think we would want this to fail entirely, but it's worth letting the user know that something is happening that they didn't expect

ketiltrout:
Yeah, that's a good compromise.

PR #241, opened 2023-04-01
MPIArray reshape fixes

jrs65:
- test(mpiarray): check more complex reshapes
- doc(mpiarray): fix docstring argument order
- fix(mpiarray): ensure reshape works properly

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -1879,11 +1879,11 @@ class MPIArray(np.ndarray):
 
         Parameters
         ----------
+        self : MPIArray or ndarray
+            The array which will be created
         obj : MPIArray, ndarray or None
             The original array being viewed or broadcast.
             When in the middle of a constructor, obj is set to None.
-        self : MPIArray or ndarray
-            The array which will be created
         """
         if obj is None:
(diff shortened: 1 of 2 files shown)


jrs65:
Gah. What kind of person asks for a review before making the CI tests pass?! 🤦

caput/mpiarray.py
```diff
@@ -1333,32 +1333,83 @@ def reshape(self, *shape):
...
+                        f"Found two wildcards (-1) in new array shape {new_shape}"
+                    )
+                else:
+                    new_size *= s
+
+            if wildcard_pos >= 0 and actual_size % new_size == 0:
+                # If there was a wildcard then it then we need to check that it would be
+                # a sane value and calculate what it is
+                new_shape = new_shape[:]
+                new_shape[wildcard_pos] = actual_size // new_size
```
ljgray, line 1365: This won't work if `new_shape` is actually a tuple, it's being given a list when this is called (line 1382). Should either change the type annotation or explicitly cast it to a list in this function.

ljgray (approved):
One quick fix but looks good otherwise, I've approved it

PR #243, opened 2023-06-09
Extra ruff rules

ljgray:
see radiocosmology/draco#243 for details on rules


ljgray:
@ketiltrout @jrs65 and here

jrs65 (approved):

PR #244, opened 2023-07-17
fix: cython fails to compile unless directives at top of file

aelanman:
The `weighted_median.pyx` cython file was failing to compile for me, because somehow cython was interpreting a bunch of objects as Python when they weren't.

I'm not sure why this only became an issue for me now, but this seems to have fixed it. I just had to move the compiler directive comments to the top of the file.

I would appreciate if anyone with more cython experience could give some insight.

### DIFF
--- a/caput/weighted_median.pyx
+++ b/caput/weighted_median.pyx
@@ -1,7 +1,8 @@
-"""Weighted Median Functions"""
-
 # distutils: language = c++
 # cython: language_level = 2
+
+"""Weighted Median Functions"""
+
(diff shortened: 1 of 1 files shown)


ljgray:
Interesting. We should hold off on merging until we understand why this fixes the issue so that we don't risk breaking anything else through this change, but this is really weird

ljgray:
I've opened a relevant issue, #245. I think that this fix should be fine but it's definitely worth understanding the root of the problem

PR #248, opened 2023-07-22
fix(runner): reset the environment to stop crash at job start

jrs65:
Before a job is submitted mpi4py modifies the environment in
a way that is invisible to Python. This modified environment is
inherited by the submitted jobs and was causing them to crash. This
change resets the environment for the subprocess that submits that job.

### DIFF
--- a/caput/scripts/runner.py
+++ b/caput/scripts/runner.py
@@ -2,6 +2,7 @@
 
 import itertools
 import os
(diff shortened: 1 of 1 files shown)


ljgray (approved):
Looks good - I assume it's been tested on both the 3.10 and 3.11 environments?

PR #253, opened 2023-08-16
parallel load with no attributes

jrs65:
- feat hints to override
- fix(memh5): error with distributed loading of files with datasets in groups
- ffirst
- refactor(memh5): move no selections sentinel to empty list to simplify

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install pip dependencies
         run: |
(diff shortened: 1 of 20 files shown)


rikvl (approved):
I think you can go ahead and merge this. I couldn't find anything wrong with this, but can't say that I have fully understood everything.

For my benefit, maybe you can explain
- In `_copy_from_file`, what would you expect the difference to be between `key` and `item.name`?
- What's the benefit of passing `fname` rather than `h5group` to `MPIArray.from_file`?

rikvl (approved):
Are the failed run-tests caused by something upstream?

PR #256, opened 2023-10-30
Avoid redistribute copy if using a single mpi process

ljgray:
This is relevant because of the check for available memory. If we are on a single node, we don't need to make a new array, so this was effectively limiting users to mpiarrays that were < ~0.4 of the available memory (which could be problematic in a notebook, for example)

### DIFF
--- a/caput/mpiarray.py
+++ b/caput/mpiarray.py
@@ -735,6 +735,13 @@ class MPIArray(np.ndarray):
         if self.axis == axis or self.comm is None:
             return self
 
+        # Avoid repeat mpi property calls
+        csize = self.comm.size
+        crank = self.comm.rank
+
+        if csize == 1:
+            return MPIArray.wrap(self.local_array, axis, self.comm)
+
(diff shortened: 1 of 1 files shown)


caput/mpiarray.py
```diff
@@ -735,6 +735,9 @@ def redistribute(self, axis: int) -> "MPIArray":
         if self.axis == axis or self.comm is None:
             return self
 
+        if self.comm.size == 1:
```
ketiltrout, line 738: If you want something pedantic, here you go:
```suggestion
        # Avoid repeat mpi property calls
        csize = self.comm.size
        crank = self.comm.rank
        
        if csize == 1:
```
(Moving lines 759-762 up here to save yourself yet another property call.)

ketiltrout (approved):
I'm assuming everything from line 763 onwards is just left-shifted one indent level.  A cursory look seems like that's true.

Otherwise, this looks fine.

PR #257, opened 2023-11-17
fix: upgrade versioneer and change to non-vendored mode for Python 3.12

jrs65:


ljgray (approved):
I think this looks fine once the merge conflict is resolved (would've been a result of merging my ruff Practice few minutes ago)

Issue #181, opened 2021-07-02
Task running order in pipeline


jrs65:
I've been finding and thinking about a few issues with the task running order within `caput.pipeline`, and I figured I would start a discussion about potential fixes or mitigations.

## Products kept around longer than needed
First let's think about an example config:
```yaml
tasks:
  - type: MakeStream
    out: stream

  - type: FilterA
    in: stream
    out: filtered_stream

 - type: FilterB
    in: stream
    out: filtered_stream

  - type: MakeMap
    in: filtered_stream
```
In this the initial `stream` is generated, each filter task is run which adds two entries to the `filtered_stream` queue and then finally the `MakeMap` tasks consumes them. While all of this is technically correct it has kept the `filtered_stream` entries around for longer than was necessary as each task is run in turn. In theory `MakeMap` could have run for the first time immediately after `FilterA` which would allowed it to have dropped the first item in `filtered_stream`. This may seem like a small issue, but we are often memory constrained and so running tasks in an order which uses more memory than required is an issue.

In the example above this could be fixed by the user by changing the task order in the file to be: MakeStream, FilterA, MakeMap, FilterB; but although this would work above, if we added a FilterC there would be no good place to place the MakeMap task.

I can see a few potential changes that would help here, all of which change add a concept of priority to the task running order which would mean they run ahead of other tasks provided their requirements are satisfied (i.e. they become greedy):

1.  prioritise tasks that have the most items in their queue already. This wouldn't help the problem above, but would definitely mean you couldn't construct an even worse version by adding more FilterX steps.
2. analyse the task graph and prioritise tasks with more connections to the input queue. This might be hard to do correctly, and would be tricky if there are different input queues. In the example above the pipeline would realise that multiple tasks produce input for MakeMap and would thus prioritise running MakeMap.
3. add a priority option to the task that allows the config writer to give specific tasks higher priority. In the example above the user would specify a higher priority for MakeMap which would mean that it is run ahead of other tasks whenever is has work to do.

## Pairing up/produce output more often than you get input

Lets say I want to stack catalog sources on a set of maps. I might have something like this:
```yaml
tasks:
  - type: LoadCatalogs
    out: catalog
    params:
      files:
      - catalogA
      - catalogB
    
  - type: LoadMaps
    out: map
    params:
      files:
      - map1
      - map2
      - map3

  - type: Stack
    in: [catalog, map]
```
[...]

Issue #152, opened 2021-01-03
Pipeline advances to "next" even if "setup" requires an input that is never generated.


jrs65:
Mostly in the title. It seems to me from my read through of some pipeline issues that if a task takes an argument for setup, and then more inputs for it's next, even if the required setup inputs aren't generated, the next will still eventually get called for it's arguments. This tends to break pipeline tasks as they don't assume they could be in this state. The correct behaviour would be to have the pipeline task finish immediately as soon as it realises it's never getting it's setup input.

@nritsche @anjakefala I've tagged one of you both in here, but we can figure out where to actually assign it later this week.

nritsche:
Do you have an example config for this by any chance?

jrs65:
Yes. But not a short one.

Here's the relevant section cut out of the daily config:
```
    # Mask out intercylinder baselines before beam forming to minimise cross
    # talk. This creates a copy of the input that shares the vis dataset (but
    # with a distinct weight dataset) to save memory
    - type: draco.analysis.flagging.MaskBaselines
      requires: manager
      in: sstream
      out: sstream_inter
      params:
        share: vis
        mask_short_ew: 1.0

    # Load the source catalogs to measure fluxes of
    - type: draco.core.io.LoadBasicCont
      out: source_catalog
      params:
        files:
        - "/project/rpp-krs/chime/chime_processed/catalogs/ps_cora_10Jy.h5"
        - "/project/rpp-krs/chime/chime_processed/catalogs/ps_CGRaBS.h5"
        - "/project/rpp-krs/chime/chime_processed/catalogs/ps_QSO_05Jy.h5"

    # Measure the observed fluxes of the point sources in the catalogs
    - type: draco.analysis.beamform.BeamFormCat
      requires: [manager, sstream_inter]
      in: source_catalog
      params:
        timetrack: 60.0
        save: true
        output_name: "sourceflux_{tag}.h5"
```

The issue is the BeamFormCat task. It can happen that no `sstream_inter` is ever produced (i.e. if there wasn't any good data to generate it). In that case the `.setup()` for the task doesn't seem to get run, but it will still advance the state and attempt to call the `.process()` for the task, which fails because `.setup()` never generated some required state.

nritsche:
Looking at `pipeline.py`, I think the pipeline gets stopped if one of the requirements is `None`: https://github.com/radiocosmology/caput/blob/b967e3eba04184ca6284e19a135fc18746f7833c/caput/pipeline.py#L981
[...]

PR #258, opened 2023-12-21
Pipeline runner improvements

ljgray:
This is a pretty involved PR, so it may take some time to review.

# Features
This PR addresses most of the issues in #181 and #152. In particular:
- The pipeline runner uses a priority system to choose the next task to run, based on
    - Number of times a task can run and its net consumption per iteration
    - Base priority as set by the user
    - Pipeline config order
- A task which is a pure producer (i.e. has no input keys) will _never_ be `available`, meaning that these will *only* run when nothing else is able to do anything
- A task will not be able to run `next` if `setup` did not run successfully
- Separate the selection of task to run next from the main run loop, which makes it easier to implement different task selection methods
- Include a `legacy` option which will mimic the current pipeline behaviour
- Add two extra task config features:
    - `limit_outputs`: restrict the number of times a task can produce output in the `next` state. Once this limit is reached, the task will immediately advance its state.
    - `base_priority`: priority modifier which is added to the calculated priority. Can be negative.

Two other features are added which are helpful for memory analysis:
1. A function to find the total memory footprint of an object, _including_ any objects it references.
2. A monitoring function in `PSUtilProfiler` which checks memory usage every 0.5 seconds in an attempt to log peak memory used by a task

# Refactoring
- Additional refactoring of parts of `pipeline.Manager` class related to initial task loading and pipeline creation. This does not change any behaviour, but slightly improves the distribution of responsibility between methods
- Update to use f-strings

# Remaining:
- [ ] Input broadcasting. This is tricky as the task would likely have to know how many total inputs to expect for each queue (i.e. how many total combinations of inputs), so it would have to wait for all inputs to accumulate instead of clearing them as soon as possible

Note: this re-implements the multi-output functionality of #263, but that has some extra tests

### DIFF
--- a/caput/pipeline.py
+++ b/caput/pipeline.py
@@ -1,4 +1,4 @@
-"""Data Analysis and Simulation Pipeline.
+r"""Data Analysis and Simulation Pipeline.
 
 A data analysis pipeline is completely specified by a YAML file that specifies
 both what tasks are to be run and the parameters that go to those tasks.
@@ -170,7 +170,7 @@ out
 requires
     A 'pipeline product key' or list of keys representing values to be passed
     as arguments to :meth:`setup`.
-in
+in\_
     A 'pipeline product key' or list of keys representing values to be passed
     as arguments to :meth:`next`.
 
@@ -180,10 +180,31 @@ for the various tasks, as specified be the 'params' keys.
 Execution Order
 ---------------
 
-When the above pipeline is executed is produces the following output.
+There are two options when choosing how to execute a pipeline: standard and legacy.
+When the above pipeline is executed in standard mode, it produces the following output.
 
 >>> local_tasks.update(globals())  # Required for interactive sessions.
->>> Manager.from_yaml_str(spam_config).run()
+>>> m = Manager.from_yaml_str(spam_config)
+>>> m.run()
+Setting up PrintEggs.
+Setting up GetEggs.
+Setting up CookEggs.
+Spam and green eggs.
+Spam and duck eggs.
+Spam and ostrich eggs.
+Finished PrintEggs.
+Cooking fried green eggs.
+Cooking fried duck eggs.
+Cooking fried ostrich eggs.
+Finished GetEggs.
+Finished CookEggs.
+
+When executed in legacy mode, it produces this output.
+
+>>> local_tasks.update(globals())  # Required for interactive sessions.
+>>> m = Manager.from_yaml_str(spam_config)
+>>> m.execution_order = "legacy"
+>>> m.run()
 Setting up PrintEggs.
 Setting up GetEggs.
 Setting up CookEggs.
@@ -197,11 +218,26 @@ Finished PrintEggs.
 Finished GetEggs.
 Finished CookEggs.
 
-The rules for execution order are as follows:
+To understand the differences, compare the rules for each strategy.
+The `standard` method uses a priority system based on the following criteria,
+in decreasing importance:
+
+1. Task must be available to execute some step.
+2. Task priority. This is set by two factors:
+
+   * Dynamic priority: tasks which have a higher net consumption
+     (inputs consumed minus outputs created).
+   * Base priority: user-configurable base priority is added to
+     the dynamic priority.
+
+3. Pipeline configuration order.
+
+If no tasks are available to run, the `legacy` method is used, which uses the
+following execution order rules:
 
 1. One of the methods `setup()`, `next()` or `finish()`, as appropriate, will
    be executed from each task, in order.
-2. If the task method is missing its input, as specified by the 'requires' or 'in'
+2. If the task method is missing its input, as specified by the 'requires' or 'in\_'
    keys, restart at the beginning of the `tasks` list.
 3. If the input to `next()` is missing and the task is at the beginning of the
    list there will be no opportunity to generate this input. Stop iterating
@@ -210,7 +246,13 @@ The rules for execution order are as follows:
 5. Once a method from the last member of the `tasks` list is executed, restart
    at the beginning of the list.
 
-If the above rules seem somewhat opaque, consider the following example which
+The difference in outputs is because `PrintEggs` will always have higher priority
+than `GetEggs`, so it will run to completion _before_ `GetEggs` starts generating
+anything. Only once `PrintEggs` is done will the other tasks run. Even though
+`CookEggs` has the highest priority, it cannot do anything without `GetEggs` running
+first.
+
+If the above `legacy` rules seem somewhat opaque, consider the following example which
 illustrates these rules in a pipeline with a slightly more non-trivial flow.
 
 >>> class DoNothing(TaskBase):
@@ -252,10 +294,12 @@ illustrates these rules in a pipeline with a slightly more non-trivial flow.
 ... no_params: {}
 ... '''
 
-The following would error, because the pipeline config is checked for errors, like an 'in' parameter without a
+The following would error, because the pipeline config is checked for errors, like an 'in\_' parameter without a
 corresponding 'out'::
 
-    Manager.from_yaml_str(new_spam_config).run()
+    m = Manager.from_yaml_str(new_spam_config)
+    m.execution_order = "legacy"
+    m.run()
 
 But this is what it would produce otherwise::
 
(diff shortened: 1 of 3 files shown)


ljgray:
I think this is ready for a review and some thoughts on the priority implementation. I've been testing it on some pipelines with good results, but I'm still thinking of edge cases where this may fail/behave worse than the old implementation, so input is helpful.

The two memory tools seem to be working as expected.

caput/pipeline.py
```diff
@@ -605,45 +643,164 @@ def run(self):
...
+
+            if task._pipeline_is_available:
+                available.append(jj)
+
+        if not available:
+            # Nothing is currently available, so fall back to a
+            # blind loop starting at the first task. If there is
+            # nothing left, this will raise StopIteration.
+            self._task_idx = self._task_head
+            return self._iter_tasks()
```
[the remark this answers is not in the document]
ketiltrout: Though I think it could still work, if you change what I suggest to put at the beginning of `_iter_tasks` above to
```
if self._task_idx is None:
    self._task_idx = self._task_head
else:
   self._task_idx += 1
```
and set `self._task_idx` to `None` here, though I don't know if that's really an improvement.

caput/pipeline.py
```diff
@@ -918,71 +1126,168 @@ def cacheable(self):
...
+        elif self._pipeline_state == "next":
+            # Calculate the total net consumption of the task
+            p = len(self._in_keys) - len(self._out_keys)
+            # How many times can the task run?
+            p *= min((q.qsize() for q in self._in), default=1)
+
+        else:
+            # If a task has passed the above states, it should be
+            # finished quickly so set a very high priority
+            p = 100
```
ketiltrout, line 1190: It's not clear to me how the value 100 compares to the values computed in the previous block.  If the intent here is to make the priority in this case be always higher that what's calculated there, I'd make this significantly (orders of magnitude) larger, just to remove any hint of ambiguity.

caput/pipeline.py
```diff
@@ -834,8 +1013,28 @@ class TaskBase(config.Reader):
...
+        NOT CURRENTLY IMPLEMENTED
+    limit_outputs : int
+        Limits the number of `next` outputs from this task before finishing.
+        Default is None, allowing an unlimited number of `next` products.
+    base_priority : int
+        Integer priority relative to other tasks. Larger positive values have
+        higher priority, and larger negative numbers have lower priority.
+        Default is 0. After setup, priority will be adjusted based on the
+        difference between number of inputs and number of outputs - tasks
+        with more input than outputs will have increased priority.
```
ketiltrout, line 1031: I think some more guidance might be helpful here.

There's two priority ordering systems playing off one another here (the `base_priority` specified by the users and the priority generated by considering inputs/outputs).  You need to explain how these systems interact with each other.

You say here that the priorities get adjusted, but you don't say how.  Possible questions:

* Is there a ballpark number for `base_priority` for a "high priority" task?  Is it 1, 10, 100, 1000?
* If I have four tasks with base priorities of 1, 2, 3, 4.  How likely is it that the tasks end up in a different priority order?  What about if I had four tasks with base priorities of 10000, 20000, 30000, 40000?
* Is the actual value of the base priority important, or only the difference between two priorities?  Or the ratio?  Would a run with tasks with base priorities (1, 2, 3) be the same as one with base priorities (-2, -1, 0), or (10, 20, 30)?

The easiest thing to do might to just have examples to for people to use as guidelines for `base_priority` for typical pipeline tasks.

ketiltrout (approved):
I don't see any real show-stoppers here.  I think the new priority system is well designed: it's both simple (to explain/implement) but flexible enough to provide users the necessary tools to adjust things when necessary.

Some extra documentation on how it works may help.

You're right that there may be some edge cases that come up in production, but I don't see anything so obvious just looking at the PR.

The legacy ordering at least means the previous behaviour can be returned to until things are fixed, if everything goes pear-shaped.  And I don't think we're locking ourselves into anything that's going to get us into trouble and we couldn't fix if needed.

I have left some comments which I think are mostly minor in scale.

caput/pipeline.py
```diff
@@ -605,45 +643,164 @@ def run(self):
...
+            return self._iter_tasks()
+
+        # Reverse sort the available tasks first by priority and second
+        # by base priority such that for any two tasks with equal priority,
+        # the task with highest base priority will be selected
+        new_index = sorted(
+            available,
+            key=lambda i: (self.tasks[i].priority, self.tasks[i].base_priority),
+            reverse=True,
+        )[0]
```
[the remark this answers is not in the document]
ljgray: I thought about it briefly, but evenly for a pipeline with a couple hundred tasks I think the extra overhead should be negligible, so I generally would favour readability over optimization

caput/pipeline.py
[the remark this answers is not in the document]
ljgray: Yeah, this is a tricky (and annoying) one, and I haven't really come up with a cleaner way of doing it. 

The thing here is that I **only** want the iterator to advance by one if I reach the point in the pipeline at line 707 *and* the method is `legacy`. In the exception handling above, I modify the task pointer and I want the next pipeline iteration to start at the new pointer location (inclusive). So, even if the execution order is `legacy`, I don't necessarily want to increment by one on every call to `_iter_tasks`.

Similarly when calling it as a fallback, I reset the task pointer to the head and I want it to include that task

caput/pipeline.py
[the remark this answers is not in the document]
ljgray: The other thing to note here is that it's only sorting the *available* tasks rather than the entire task list, and generally only a handful of tasks will be available at any given time

caput/pipeline.py
ljgray: I've updated the documentation to try to be clearer and more explicit. The only thing I'm debating is whether `base_priority` should be an absolute - i.e., the `priority` method returns just the `base_priority` if it's set, or to continue with this method of combining base and dynamic priority

caput/pipeline.py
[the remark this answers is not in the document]
ketiltrout: I think it's best to keep the code as simple as you have it now, at least until the optimisation becomes important.

caput/pipeline.py
ketiltrout: The updated description is great.

I think it's better to keep the system as it is, which gives flexibility to use `base_priority` either as an (effectively) absolute priority (by using a large number), or else as a priority hint for the automatic priority calculation (by using a small number).

caput/pipeline.py
```diff
@@ -650,76 +818,95 @@ def _check_task_output(out, task):
...
+                    location=task_spec if e.line is None else e.line,
                 ) from e
 
-    @staticmethod
-    def _validate_task(task, in_, requires, all_out_values):
-        # Make sure this tasks in/requires values have corresponding out keys from another task
-        for key, value in (["in", in_], ["requires", requires]):
-            if value is not None:
+    def _validate_task_inputs(self):
+        # Make sure all taskss in/requires values have corresponding
```
ketiltrout, line 873:
```suggestion
        # Make sure all tasks' in/requires values have corresponding
```
typo

caput/pipeline.py
```diff
@@ -765,55 +953,77 @@ def _setup_task(self, task_spec):
...
-            k: v for k, v in task_spec.items() if k in ["requires", "in", "out"]
-        }
-
         # Create and configure the task instance
         try:
             task = task_cls._from_config(task_params)
         except config.CaputConfigError as e:
             raise config.CaputConfigError(
-                f"Failed instantiating {task_cls} from config:\n\t{e}",
+                f"Failed instantiating {task_cls} from config./n/t{traceback.format_exc()}",
```
ketiltrout, line 969:
```suggestion
                f"Failed instantiating {task_cls} from config.\n\t{traceback.format_exc()}",
```
typo

ketiltrout (approved):
This looks reasonable.  It's difficult, with static analysis, to anticipate what sort of corner cases (if any) we might get ourselves into with this change, but I think this is implemented well enough that's it's a good place to start.

I found a couple of typos:

caput/pipeline.py
ljgray: Oh that would've been gross

PR #259, opened 2024-01-17
Enforce linting rule RUF012

ljgray:
See radiocosmology/draco#258

### DIFF
--- a/caput/pipeline.py
+++ b/caput/pipeline.py
@@ -625,8 +625,8 @@ class Manager(config.Reader):
                     except _PipelineMissingData:
                         if self.tasks.index(task) == 0:
                             msg = (
-                                "%s missing input data and is at beginning of"
-                                " task list. Advancing state." % task.__class__.__name__
+                                f"{task.__class__.__name__} missing input data and is at beginning of"
+                                " task list. Advancing state."
                             )
                             logger.debug(msg)
                             task._pipeline_advance_state()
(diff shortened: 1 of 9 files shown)


caput/pipeline.py
```diff
@@ -750,11 +744,7 @@ def _setup_task(self, task_spec):
             try:
                 task_cls = misc.import_class(task_path)
             except (config.CaputConfigError, AttributeError, ModuleNotFoundError) as e:
-                msg = "Loading task '{}' caused error {}:\n\t{}".format(
-                    task_path,
-                    e.__class__.__name__,
-                    str(e),
-                )
+                msg = f"Loading task '{task_path}' caused error {e.__class__.__name__}:\n\t{e!s}"
```
ketiltrout, line 747: What's the reason for embedding `e` in the re-raised `CaputConfigError` here (I'm mostly talking about the `\t{e!s}` part)?

`raise BarError from FooError` is already going to report the original `FooError` in one of its two tracebacks.

And if something else is catching the `CaputConfigError`, isn't it better for _that_ to get the original `e` out of `CaputConfigError.__cause__` if it wants it?

ketiltrout (approved):
My comment above is mostly aesthetic.  I don't think it's harmful as-is.

caput/pipeline.py
ljgray: Honestly? I'm not sure. I think it was just written this way when the original pipeline runner was put together years and years ago.

I have another branch where I'm making improvements to the pipeline runner, so I'll make some error message improvements there instead so that this branch only makes style changes

ketiltrout: Fair.

PR #263, opened 2024-02-06
Pipeline tasks can return multiple outputs to unpack

jrs65:
- test(pipeline): ensure multiple output pipeline tasks are supported
- fix(pipeline): pipeline tasks can return multiple outputs to unpack

### DIFF
--- a/caput/tests/conftest.py
+++ b/caput/tests/conftest.py
@@ -87,7 +87,7 @@ class CookEggs(IterBase):
 
     def process(self, _input):
         """Run process."""
-        print("Cooking %s %s eggs." % (self.style, input))
+        print("Cooking %s %s eggs." % (self.style, _input))
 
     def finish(self):
         """Run finish."""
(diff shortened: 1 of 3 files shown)


caput/tests/conftest.py
```diff
@@ -124,6 +124,25 @@ def write_output(self, filename, output, file_format=None, **kwargs):
...
+multi_eggs_pipeline_conf = """
+---
+pipeline:
+  tasks:
+    - type: caput.tests.conftest.GetEggs
+      params: eggs_params
+      out: [color, egg]
+    - type: caput.tests.conftest.CookEggs
+      params: cook_params
+      in: egg
```
rikvl, line 136:
```suggestion
      in: color
```
Not 100% sure if I understood the intention of this test. Were the same eggs supposed to be fried twice? If the idea was to test a task with multiple outputs, where the outputs get used by subsequent tasks, then I think `color` should be an input somewhere.

caput/tests/test_pipeline.py
```diff
@@ -8,3 +8,10 @@ def test_pipeline():
     result = conftest.run_pipeline()
     print(result.output)
     assert result.exit_code == 0
+
+
+def test_pipeline_multiple_outputs():
+    """Test running a very simple pipeline."""
```
rikvl, line 14:
```suggestion
    """Test running a very simple pipeline with a multiple-outputs task."""
```

rikvl (changes requested):
Hope I understood what's going on here. If so, I think this only requires one or two small changes.

ljgray:
The key functionality here was added in a different PR, so this is now just adding tests

ketiltrout (approved):
This seems fine.

Though, `import`ing `conftest` directly, instead of just setting up a bunch of fixtures is a weird way of doing things.  But I don't think this PR is the place to address that.

PR #265, opened 2024-05-31
feat(pipeline): better handling for key rejection

ljgray:


caput/pipeline.py
```diff
@@ -510,6 +511,11 @@ class Manager(config.Reader):
         through all tasks based on the config order, and tries to clear out finished
         tasks as soon as possible. `standard` uses a priority and availability system
         to select the next task to run, and falls back to `legacy` if nothing is available.
+    key_pattern : str, optional
+        Regex pattern to match on in order to pass a key to subsequent tasks. By default,
+        any key that contains no alphanumeric characters (except _) will be ignored.
+        This is useful for controlling which keys are passed in tasks which produce
+        multiple outputs. Default is `[^\W_]`
```
ketiltrout, line 518:
```suggestion
    key_pattern : str, optional
        Regex pattern to match on in order to pass a key to subsequent tasks. 
        This is useful for controlling which keys are passed in tasks which produce
        multiple outputs. Default is `[^\W_]`, which will cause any key that contains
        no alphanumeric characters to be ignored.
```
You need to be more explicit that the default behaviour you describe is a result of the default value for this parameter.

Also `_` is already not an alphanumeric character, calling it out is confusing.

If you want to keep the actual default value at the end of the block like the rest of the parameter descriptions, you could do:
> Default, which will cause [...] to be ignored, is `[^\W_]`

ketiltrout (approved):
This looks fine, though I'm being pedantic in the docstrings again...

caput/pipeline.py
ljgray: I think the reason I called it out is because `\w` will usually match on `_`, but by changing the doctoring I don't think it's needed to mention that

ketiltrout (approved):

PR #266, opened 2024-06-17
Pin numpy < 2

ljgray:
At the moment, there are no available bitshuffle wheels built for `numpy>=2`. Also, our `weighted_median` module has some compatibility issues with `numpy >= 2`. For now I've just pinned `numpy < 2` until I can figure out these incompatibilities

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ cachetools
 click
 cython
 h5py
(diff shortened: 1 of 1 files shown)


ketiltrout (approved):
This is prudent until we've fully investigated the impact of numpy-2.

PR #267, opened 2024-06-24
Increase minimum python to 3.9

ljgray:
Python 3.8 will reach end-of-life in October 2024, so we should increase the minimum supported version


ketiltrout (approved):
I've updated the branch protection rules here on GitHub to accommodate the workflow changes here.

PR #269, opened 2024-07-16
feat(interferometry): Import from ch_util and driftscan

ketiltrout:
This imports two general-purpose angle-related functions from other modules using caput, with the intention of re-exporting them back to those modules:

* `rotate_ypr`: rotates a 3-basis by a yaw, pitch, and roll. Originally from drifscan: https://github.com/radiocosmology/driftscan/blob/8def16f7d1f9d35f729305dd8998c40a7cb7aceb/drift/telescope/cylbeam.py#L44
* `sphdist`: calculates the angular distance betwee two points on the sphere.  Originally from ch_util: https://github.com/chime-experiment/ch_util/blob/c460b37b27f5bcaa1dfbe7897f3ade88e6e6775b/ch_util/ephemeris.py#L358

I have made some minor changes to the docstrings.  There are also some very rudimentary unit tests.

### DIFF
--- a/caput/tests/test_interferometry.py
+++ b/caput/tests/test_interferometry.py
@@ -1,0 +1,61 @@
+"""Test interferometry routines."""
+
+import pytest
+from math import pi, sqrt
+
+from caput import interferometry
+
+
+def test_sphdist():
+    from skyfield.units import Angle
+
+    # 90 degrees
+    assert interferometry.sphdist(
+        Angle(radians=0), Angle(radians=0), Angle(radians=pi / 2), Angle(radians=0)
+    ).radians == pytest.approx(pi / 2)
+    assert interferometry.sphdist(
+        Angle(radians=0), Angle(radians=0), Angle(radians=0), Angle(radians=pi / 2)
+    ).radians == pytest.approx(pi / 2)
+
+    # 60 degrees
+    assert interferometry.sphdist(
+        Angle(radians=0), Angle(radians=0), Angle(radians=pi / 4), Angle(radians=pi / 4)
+    ).radians == pytest.approx(pi / 3)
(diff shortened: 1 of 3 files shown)


ketiltrout:
I guess there was a ruff update?  Do you want me to correct this linting error here?

rikvl:
> I guess there was a ruff update? Do you want me to correct this linting error here?

Yes, I think I'd like to have it corrected here.

Is the reason not to correct it here to keep the code added to `caput` by this PR identical to the code removed from `ch_util`?

rikvl:
> I guess there was a ruff update? Do you want me to correct this linting error here?

Maybe, yes, in a separate commit and not squash the commits?

caput/tests/test_interferometry.py
```diff
@@ -0,0 +1,61 @@
...
+    z0 = 0.5
+    basis = interferometry.rotate_ypr([pi / 2, pi / 3, pi / 4], x0, y0, z0)
+
+    # The calculation goes like this:
+
+    # After yawing by 90 degrees:
+    # x1 = -y0               y1 = x0                     z1 = z0
+
+    # Then pitching by 60 degrees:
+    # x2 = x1                y2 = 0.5 * y1 + v3/2 z1     z2 = -v3)/2 y1 + 0.5 z1
```
rikvl, line 54: I guess `v2` and `v3` refer to sqrt(2) and sqrt(3)? For consistency, maybe remove the `)` after the second `v3`?

ljgray (approved):

ljgray:
> I guess there was a ruff update?  Do you want me to correct this linting error here?

Yeah, given that it's a small error it's probably fine to fix it here.

PR #270, opened 2024-07-24
fix(time): Make Star.names iterable in skyfield_star_from_ra_dec

ketiltrout:
The skyfield documentation is poor, but I think the intention is that the `Star.names` attribute is meant to be iterable.  The skyfield code doesn't reference `Star.names` at all, and there's no documentation on the meaning of the attribute (or any of the parameters of `Star.__init__`), so it's very much unclear.

The default value for `Star.names` is an empty tuple.  Probably any (non-`str`) iterable would be reasonable.

This changes `skyfield_star_from_ra_dec` to convert a `str` passed as `name` into a single-element tuple when assigning it to `Star.names`, and also updates the docstring to indicate that a tuple or list of names is an acceptable value for the `name` parameter.

### DIFF
--- a/caput/time.py
+++ b/caput/time.py
@@ -64,6 +64,10 @@ determined by the following (in order):
 - By setting the environment variable ``CAPUT_SKYFIELD_PATH``
 - If neither of the above is set, the data is place in ``<path to caput>/caput/data/``
 
+Other skyfield helper functions:
+
+- :py:meth:`skyfield_star_from_ra_dec`
+
 Constants
 =========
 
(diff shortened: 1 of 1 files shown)


ljgray (approved):
I think these changes are fine. I'm not 100% clear on what skyfield intends here - is the idea to have a single body which can have multiple names?

ketiltrout:
>I'm not 100% clear on what skyfield intends here

I'm with you here.  Skyfield needs better docs, but my guess is it's meant to handle the fact that stellar bodies all have multiple names.  e.g. there's 50-ish alternate names listed for Cas-A in our own source catalog in `ch_util`.

ketiltrout:
This comes out of me thinking about this old `ch_util` PR: https://github.com/chime-experiment/ch_util/pull/24

PR #276, opened 2024-08-08
Numpy 2.0 support

ljgray:
- compiles against numpy >= 2.0.0rc1
- calls `np.import_array()` in `weighted_median` to include required headers
- replaces deprecated `datetime.datetime.utcfromtimestamp` with `datetime.datetime.fromtimestamp(..., datetime.timezone.utc)`
- Uses legacy numpy print options - in numpy 2, numpy scalars are printed with type information, i.e. `np.float64(3.0)` instead of `3.0`. This breaks doctests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,8 +26,7 @@ jobs:
 
       - name: Install pip dependencies
         run: |
-          pip install -e .
-          pip install -e .[lint]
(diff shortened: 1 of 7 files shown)


ketiltrout (approved):
The `set_printoptions` thing is annoying, but probably the best solution in this case.

PR #285, opened 2024-12-18
Properly propagate error in `bit_truncate_relative`

ljgray:
Negative values would end up trying to truncate against negative error, which would always return `-0.0`. This makes sure that the error value is positive.


ssiegelx (approved):
Nice catch

PR #289, opened 2025-01-30
fix(fileformats): fix compression check

ssiegelx:
Do not check against H5Filter if H5Filter is None.

### DIFF
--- a/caput/fileformats.py
+++ b/caput/fileformats.py
@@ -88,7 +88,9 @@ class HDF5(FileFormat):
         super(HDF5, HDF5).compression_kwargs(compression, compression_opts, compressor)
         if compressor:
             raise NotImplementedError
-        if compression in ("bitshuffle", H5FILTER, str(H5FILTER)):
+        if (compression == "bitshuffle") or (
+            (H5FILTER is not None) and compression in (H5FILTER, str(H5FILTER))
+        ):
             if not compression_enabled:
                 raise ValueError(
                     "Install with 'compression' extra_require to use bitshuffle/numcodecs compression filters."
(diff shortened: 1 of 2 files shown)


ssiegelx:
If the bitshuffle filters are not installed, `H5FILTER` is set to `None`.  
However, when `HDF5.compression_kwargs()` is called with `compression=None`, it incorrectly matches against `H5FILTER`, leading to an attempt to use bitshuffle compression even though it is not installed.

This issue was encountered when installing `caput` without compression support and attempting to load existing HDF5 files. The fix ensures that compression is only applied when explicitly available.

caput/fileformats.py
```diff
@@ -88,7 +88,9 @@ def compression_kwargs(compression=None, compression_opts=None, compressor=None)
         super(HDF5, HDF5).compression_kwargs(compression, compression_opts, compressor)
         if compressor:
             raise NotImplementedError
-        if compression in ("bitshuffle", H5FILTER, str(H5FILTER)):
+        if (compression == "bitshuffle") or (
+            (H5FILTER is not None) and compression in (H5FILTER, str(H5FILTER))
+        ):
```
ketiltrout, line 93: Given that whenever `H5FILTER` is set to `None` (up on Line 27), then `compression_enabled` is also set to `False`, and recognising that the only value of `compression` that can cause `ValueError` to get raised is the `"bitshuffle"` literal, it might be more legible to pull the `raise ValueError` check that happens immediately after this `if` out of the `if`-block and turn this into something like:
```suggestion
        if compression == "bitshuffle" and not compression_enabled:
            raise ValueError(
                "Install with 'compression' extra_require to use bitshuffle/numcodecs compression filters."
            )

        if compression_enabled and compression in ("bitshuffle", H5FILTER, str(H5FILTER)):
```
(Also need to delete the next four lines containing the `if not compression_enabled` block.)

ketiltrout (approved):
Your fix looks fine as-is, but here's a proposal for perhaps a tidier way to do it.

caput/fileformats.py
ssiegelx: Thanks @ketiltrout!  I've implemented this suggestion.

PR #290, opened 2025-03-04
feat(time): add object_coords and cirs_radec method to Observer class

ssiegelx:


ssiegelx:
`object_coords` is the recommended method for determining source coordinates and is widely used in the calibration code.  We would like to move calibration from `ch_pipeline` to `draco` and generalize it for arbitrary driftscan telescopes. 

This PR moves both `object_coords` and `cirs_radec` from `ch_ephem` to `caput`, adding them as methods of the `Observer` class.

A corresponding PR will add an alias in `ch_util.ephemeris` after this is merged.

caput/time.py
```diff
@@ -818,6 +818,85 @@ def lunar_rising(self, t0, t1=None, step=None):
...
+            Skyfield Star object with positions in ICRS coordinates
+        """
+
+        from skyfield.functions import to_polar
+
+        ts = skyfield_wrapper.timescale
+
+        epoch = ts.tt_jd(np.median(source.epoch))
+
+        pos = obs.skyfield_obs().at(epoch).observe(source)
```
ketiltrout, line 841:
```suggestion
        pos = self.skyfield_obs().at(epoch).observe(source)
```

caput/time.py
```diff
@@ -818,6 +818,85 @@ def lunar_rising(self, t0, t1=None, step=None):
...
+        deg : bool
+            Return coordinates in degrees if True, radians if False (default).
+
+        Returns
+        -------
+        ra, dec: float
+            Position of the source.
+        """
+        if date is None:  # No date, get ICRS coords
+            if isinstance(body, Star):
```
ketiltrout, line 880:
```suggestion
            if isinstance(source, Star):
```
et seq.

ketiltrout (changes requested):
I think this is great.   I'm always for moving stuff from ch_ephem to caput, though I think there's been some trouble with renaming the parameters when these were ported.

What about `ch_ephem.coord.star_cirs`?  Should we move that one too?  It's basically just some boilerplate around `cirs_radec`.

caput/time.py
```diff
@@ -818,6 +818,85 @@ def lunar_rising(self, t0, t1=None, step=None):
...
+                ra, dec = body.ra.radians, body.dec.radians
```
mwilensky768, line 881:
```suggestion
                ra, dec = source.ra.radians, source.dec.radians
```

mwilensky768 (changes requested):
I found one other line where the `body`-> `source` rename didn't happen. As far as I could tell from inspection, it looks like there shouldn't raise `NameError` once those changes are made, and I didn't see anything go missing going from `ch_ephem` to here (except what ought to have gone missing based on the fact that this is now a method on `Observer` so obs -> self etc.). 

If this is agreeable I'm happy to approve.

ssiegelx:
Thanks! I'm not sure how I missed renaming those variables when porting the code over, but that's fixed now. I've also moved over the `star_cirs` method.

ketiltrout (approved):
Looks good.

mwilensky768 (approved):

PR #291, opened 2025-03-10
Import pyfftw before numpy

ljgray:
There's a bug in `pyfftw` which seems to cause some sort of namespace issues with `numpy`. It seems like importing `pyfftw` before `numpy` fixes the issue, but this is environment-wide (i.e., it probably won't work in complex pipeline code where numpy is usually one of the first imports). However, I don't see any issue in at least slightly improving peoples chances of having it work.

This bug seems to still exist in the most recent version of `pyfftw`, so hopefully it gets fixed at some point.

Also, some fixes for ruff.

### DIFF
--- a/caput/fftw.py
+++ b/caput/fftw.py
@@ -49,16 +49,21 @@ Functions
 
 from __future__ import annotations
 
-import numpy as np
-
-from caput import mpiutil
-
+# NOTE: Due to a bug in pyfftw, it needs to be imported before
+# numpy in order to avoid some sort of namespace collision.
+# If you run into a RuntimeError when trying to use the `FFT`
+# class, make sure that your environment imports `pyfftw`
+# before `numpy`. Hopefully this will be fixed soon.
 try:
     import pyfftw
 except ImportError as exc:
     raise ImportError(
         "`pyfftw` is not installed. Install `pyfftw` via `caput[fftw]`."
     ) from exc
+
+import numpy as np
+
+from caput import mpiutil
 
 
 class FFT:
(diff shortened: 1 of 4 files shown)


ketiltrout (approved):
This is unfortunate.

Have you tried deleting `numpy` from `sys.modules` before attempting to load `pyfftw`?

I can't seem to reproduce this error (or maybe I'm not understanding it) but does a contrived example like:

```
import numpy as np
del sys.modules["numpy"]
import pyfftw
```
work?

`sys.modules` isn't actually the list of imported modules, just a Python-level-accessible version of it, so it doesn't actually unload the module, and it doesn't affect the local `np` var which references the numpy module in the local scope, but I think it will make it appear to the `pyfftw` internal Python code as if numpy wasn't loaded.

I agree that this difficult to fix project-wide, but there's nothing wrong with the work-around you're using here.

Have you reached out to the pyfftw devs for advice?

ljgray:
Yeah, it would help if I had actually explained the issue correctly, or even at all...

It isn't actually `pyfftw` and `numpy`, but rather the underlying `fftw` library and (it appears) some `MKL` bindings for ffts which conflict. `fftw` fails to create a path if dimension of the dataset minus the number of axes being transformed is >= 2. 

This bug doesn't show up running on my laptop, but it does happen on Cedar (no MKL on my macbook is the guess).

ketiltrout:
So it's an issue with the ordering of clib loading for extensions? 😬   Yeah, that's going to be tough to deal with in general.  Hmm.

ljgray:
Yeah. One option (mentioned [here](https://github.com/pyFFTW/pyFFTW/issues/40)) is to compile `pyfftw` ourselves, but I'm not convinced I want to do that. Interestingly, this thread is from 2014, but the issue still shows up in the most recent releases of `pyfftw`.

PR #292, opened 2025-03-12
Ljg/py310

ljgray:


ketiltrout:
First off: I don't think `caput` need by hamstrung by what's going on with CHIME, per se.

That said, however, from a practical point of view, we should at least consider how this will affect (re-)deployment of CHIME software to the acquisition system.  Do we know where we're using `caput` on the telescope side and how out-of-date the Python versions we might have there?

I guess anywhere we have `ch_util` installed would need caput...

ljgray:
Good point. We could have a discussion on a call where more people can weigh in. 

I guess this will always be an issue, so it would be good to come up with some sort of documented plan.

ketiltrout (approved):
I took a look at `ch_ansible` to see where we're installing `ch_util` and then looked at those machines.

My survey is as follows:
* `wind`: 3.10 -- though alpenhorn now requires 3.11 so this will need an upgrade anyways
* `recv1/2`: 3.8
* `marimba`: 3.8
* `agogo`: 3.8
* `hk-east/west`: 3.5
* `carillon`: 3.7

So, given the current minimum here is 3.9, I think we can say the ship has already sailed, and this isn't making things worse.

ketiltrout:
It's also important to note that we can't run the CI on 3.8 anymore (because it's out of support from both GitHub and the python devs themselves), so keeping support for that isn't really an good option.

PR #295, opened 2025-04-07
Add a flag to deprecate config properties

ljgray:
This will be helpful when modifying old tasks.

I've also added a feature to the pipeline base task to add a debug log entry if a task returns the same object as one of the inputs, which suggests that it could have been modified in place. This is mostly for my own use if I'm trying to review weird behaviour in a pipeline config, but I think it's helpful.

There's something broken in the tests from a very recent release of `numcodecs` (https://numcodecs.readthedocs.io/en/stable/release.html#removals). 

There's already an open PR in `zarr` to address it, so I'm just going to leave this PR until that gets merged, because I don't want to have to pin `numcodecs` ourselves. Update: this still wasn't working properly with python 3.13, so I've just pinned `numcodecs`. I think we'll have to keep it pinned until we upgrade to zarr 3 anyway.

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -58,6 +58,7 @@ if TYPE_CHECKING:
     from . import fileformats
 
 import logging
+import warnings
 
 from yaml.loader import SafeLoader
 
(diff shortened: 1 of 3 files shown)


ljgray:
See https://github.com/zarr-developers/zarr-python/pull/2965

ketiltrout (approved):

PR #296, opened 2025-04-17
Try to warn user about specific key in deprecation warning

ljgray:
Right now the output is something like 
```
UserWarning: Property <caput.config.Property object at 0x14d8eb2ae580> is deprecated and may behave unpredictably.
```

which isn't very helpful. We should at least try to tell the user something about the name of the property.

Unfortunately, with the way that `config.Property` is typically used, the key won't get set until the first time that `__get__` or `__set__` is called. This just means that if someone is deprecating a property, they should also provide an argument for `key` when they set `deprecated=True`.

Or, we could delay this warning until the first time `__get__` or `__set__` is called. Thoughts?

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -87,17 +87,13 @@ class Property:
             An easy way to flag a property as deprecated. This will raise a deprecation
             warning, but will not change the behaviour of this `Property` instance.
         """
-        if deprecated:
-            warnings.warn(
-                f"Property {self!r} is deprecated and may behave unpredictably. "
-                "Check the documentation of the class/function where this is being "
-                "initialised to see the recommended solution."
-            )
-
         self.proptype = (lambda x: x) if proptype is None else proptype
         self.default = default
         self.key = key
         self.propname = None
+        # If this property is deprecated, we'll emit a warning
+        # when it's used
+        self._deprecated = deprecated
 
     def __get__(self, obj, objtype):
         # Object getter.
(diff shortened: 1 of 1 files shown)


ketiltrout:
As-is, when is the warning issued?  When the module containing the class definition is imported?  Or when the class is instantiated?

ljgray:
Yeah, when the warning is issued. At the moment it gets issued when you first import the module because that's when the `Property` class gets instantiated, which is nice because you'll get the warning if you lint a pipeline config, for example.

ketiltrout:
Hrm.

That's maybe good for _maintainers_ of the project, but it's not great for the code's _users_. If I get a deprecation warning even if I'm not using the deprecated property, then there's no way for me (a user of the code) to get rid of the warning other than haranguing the maintainer of the code to remove the deprecated property.

Which I think means I'm arguing that the deprecation warning should be moved to `__get__`/`__set__`.

In a codebase meant for public consumption, marking old features as deprecated should be encouraged, because that's how you allow for backwards compatibility. An API simply _having_ deprecated features shouldn't be cause for a warning.  Only if I, a user of the API, try to access the deprecated parts should I see a warning.  Emitting a warning whenever deprecated Properties exist just discourages marking Properties as deprecated.

The deprecation warning should be emitted when downstream code references the deprecated property, because that's the code that needs to be updated to resolve the deprecation.  It's of no help for the warning to occur when the deprecated property is defined in the upstream code.  The maintainer of _that_ code already knows the property is deprecated.  That's why they put `deprecated=True` in the Property instantiation.

ketiltrout (approved):
That all said, this change is an improvement on the current behaviour.  I see no problem with merging, if you'd prefer kicking the other stuff down the road.

ljgray:
Eh, I think what you said makes sense and it looks like it should be an easy adjustment, so I'll do that

PR #297, opened 2025-04-25
Make sure deprecation warning doesn't show up when loading from config unless the property is set

ljgray:
The last change generally worked, except that it would emit a warning when linting a config file since all properties were checked, even those that weren't set. This fixes that, so now you'll only get a warning when linting if you're actually setting the property in your config file.

### DIFF
--- a/caput/config.py
+++ b/caput/config.py
@@ -91,8 +91,6 @@ class Property:
         self.default = default
         self.key = key
         self.propname = None
-        # If this property is deprecated, we'll emit a warning
-        # when it's used
         self._deprecated = deprecated
 
     def __get__(self, obj, objtype):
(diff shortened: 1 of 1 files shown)


caput/config.py
```diff
@@ -91,17 +91,16 @@ def __init__(self, default=None, proptype=None, key=None, deprecated=False):
         self.default = default
         self.key = key
         self.propname = None
-        # If this property is deprecated, we'll emit a warning
-        # when it's used
-        self._deprecated = deprecated
+        self.__deprecated = deprecated
```
ketiltrout, line 94: Why the double underscore?  Do you want it mangled?

ketiltrout (approved):
This seems fine.

caput/config.py
ljgray: I was going back and forth on this. I had a weak argument that mangling could be useful since the user should never be trying to access the value of `deprecated` so why not, but it's also more pythonic to just make it private and I don't really think there's much of a case for extending this class in a way that could conflict with that attribute. So I can just revert that.

PR #298, opened 2025-05-08
(1/?) - move pipeline infrastructure and assorted utilities into `caput`

ljgray:
This is the first PR of several.

This PR moves all the pipeline infrastructure from draco to caput (`task` and `containers`), and includes a few basic, fundamental tasks (pipeline flow, basic I/O, etc...). It also moves some general utilities from elsewhere, such as the `coordinate` transforms found in cora.

This PR does not attempt to significantly organize or rework anything, which will be done in future PRs. I started trying to organize the `containers` and `task` modules, but I'm not sure if those will stay the way they are. I'm very open to thoughts.


pyproject.toml
```diff
@@ -101,5 +101,6 @@ exclude = [
     "build",
     "doc",
     "tests",
+    "*/__init__.py",
```
ketiltrout, line 104: The alpenhorn solution here is:
```
[tool.ruff.lint.per-file-ignores]
"alpenhorn/*/__init__.py" = ["I", "F401"]
```

You may be able to do something similar to better fine-tune the disabled tests for `__init__.py` files.

caput/task/io.py
```diff
@@ -0,0 +1,1018 @@
...
+                raise ConfigError(f"File not found: {files!s}")
+
+        return [files]
+
+    raise ConfigError(
+        f"Argument must be list, glob pattern, or file path, got {files!r}"
+    )
+
+
+def list_of_filegroups(groups: list[dict] | dict) -> list[dict]:
```
ketiltrout, line 132: As with `list_of_filelists`, is this used anywhere other than `draco.core.io`?  Is there a reason not to just leave it there?

caput/containers/_basic.py
```diff
@@ -0,0 +1,45 @@
+"""Basic caput containers."""
```
ketiltrout, line 1: See also my comment in the cora PR about `FreqContainer`.

ketiltrout (approved):
Overall, this refactor is definitely an improvement, though there's certainly some rough edges to smooth out.  I don't see anything show-stoppy.

caput/task/io.py
ketiltrout: Ah, It's used by `ch_pipeline`.

caput/coordinates/sphcoords.py
```diff
@@ -1,4 +1,8 @@
-"""Useful functions for radio interferometry.
+"""Useful functions for dealing with spherical coordinates.
```
ketiltrout, line 1: Looking through the `ch_pipeline` PR, I think this should be renamed `caput.coordinates.spherical`.  You don't need the `coords` in the name here; that's already provided by the parent module.

caput/task/_core.py
```diff
@@ -0,0 +1,730 @@
+"""An improved base task implementing easy (and explicit) saving of outputs."""
```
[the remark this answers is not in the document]
ljgray: Updated, but will improve these docstrings later on

caput/task/io.py
ljgray: Yeah, these could maybe go somewhere else but I'm not really sure what the best place for them is.

PR #299, opened 2025-05-27
Make a `fft` module, which contains the existing `fftw` module

ljgray:
I want an easy way for users to access faster FFTs than what's provided by `numpy.fft` (which is quite slow). I had previously tried to do that with the `fftw` module, which is faster than `numpy/scipy`, but it ended up being finicky and has some annoying bugs. `fftw` is now accessed as a submodule of the `fft` submodule.

At the moment, the main `fft` module just re-exports `scipy.fft` with multiple workers set by default (`scipy` uses `pocketfft` as a backend, so I actually change the number of workers there). A full benchmark would probably be useful, but I've found that the overhead of using multiple workers is nullified quite quickly by the performance gained as arrays grow in size.

Going forward, I want to see if I can find some other, faster options. It should be reasonable to use `mkl`, and [kfr](https://github.com/kfrlib/kfr) seems interesting and has been used by LIGO, but I don't think it has python bindings. I also need to investigate options for AMD cpus, since that's what `fir` will be using.

### DIFF
--- a/caput/fft/__init__.py
+++ b/caput/fft/__init__.py
@@ -1,0 +1,12 @@
+"""caput.fft.
+
+Utilities for FFTs.
+
+Features
+========
+- Multi-threaded `scipy.fft` by default
+- Helpers to use `pyFFTW`
+"""
+
+from ._scipy_fft import *  # noqa: F403
+from . import fftw as fftw
--- a/caput/fft/_scipy_fft.py
+++ b/caput/fft/_scipy_fft.py
@@ -1,0 +1,36 @@
+"""Scipy FFT routines.
+
+This module just re-exports `scipy.fft`, but it uses
+all cores by default (instead of one). Users can still
+set the number of cores on a case-by-case basis using
+the standard `scipy` syntax.
+
+For small arrays, it will generally be faster to use a
+single threaded version from `scipy.fft` directly.
+"""
+
+import scipy.fft
+
+from .. import mpiutil
+
+# Overwrite any existing backends on import
+scipy.fft.set_global_backend("scipy", only=True)
+
+# Re-export only from `scipy.fft`
+__all__ = scipy.fft.__all__
+
+# Only use this lookup once
+_nworkers = mpiutil.cpu_count()
+
+
+def _set_workers(func):
+    def _inner(*args, **kwargs):
+        with scipy.fft.set_workers(kwargs.pop("workers", _nworkers)):
+            return func(*args, **kwargs)
+
+    return _inner
+
+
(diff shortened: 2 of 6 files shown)


caput/fft/_scipy_fft.py
```diff
@@ -0,0 +1,25 @@
...
+from .. import mpiutil
+
+# Re-export only from `scipy.fft`
+__all__ = scipy.fft.__all__
+
+# Overwrite any existing backends on import
+scipy.fft.set_global_backend("scipy", only=True)
+# Set the default number of workers
+# NOTE: I'm not sure if this could break anything. Be careful
+scipy.fft._pocketfft.helper._config.default_workers = mpiutil.cpu_count()
```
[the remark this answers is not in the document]
ljgray: Unfortunately this doesn't seem to work - the `set_workers` context manager doesn't seem to stay active when the returned function is called. However, I agree that using nested private attributes isn't a good idea, so I'll see if I can come up with something else.

caput/fft/_scipy_fft.py
```diff
@@ -0,0 +1,36 @@
...
+    def _inner(*args, **kwargs):
+        with scipy.fft.set_workers(kwargs.pop("workers", _nworkers)):
+            return func(*args, **kwargs)
+
+    return _inner
+
+
+# Re-export all scipy symbols with multiple workers set
+def __getattr__(name):
+    return _set_workers(getattr(scipy.fft, name))
```
ketiltrout, line 36: Ah, yeah.  This is what I was trying to figure out how to do: create a decorator for all the `scipy.fft` functions without having to manually decorate each of them.

caput/fft/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+Utilities for FFTs.
+
+Features
+========
+- Multi-threaded `scipy.fft` by default
+- Helpers to use `pyFFTW`
+"""
+
+from ._scipy_fft import *  # noqa: F403
+from . import _fftw as fftw
```
ketiltrout, line 12: If we're doing this, is there any reason to call the file `_fftw.py` instead of just `fftw.py`?

ketiltrout (approved):
I think this is probably a good idea.  Providing easier ways to access performant FFT implementations is good.

caput/fft/__init__.py
ljgray: Nah I don't think there's any reason not to just call it `fftw.py`

ljgray:
As mentioned, I want to try to add more performant backends than just scipy, so this is just an interim PR to set up the module structure.

PR #301, opened 2025-06-21
Move memh5, mpiarray, and related to a new submodule

ljgray:
I'm fairly keen for feedback on this one. In particular, I'm trying to figure out the best way to group memh5/mpiarray and other things that are probably relevant. For example, I've included `truncate` here, but I haven't included `mpiutil` because it seems like that has wider use outside of just the memh5 stuff.


ketiltrout:
Yeah, I'm not either, but I don't have a great alternative.

What about just `data`?  Is that too broad?

Then we'd have
```
caput.data.mem5h
caput.data.tod
caput.data.fileformats
etc.
```

I need to look at this PR more deeply...

ljgray:
Ok, I've finally got the tests working here. 

This PR still doesn't feel 100% complete, but I think I have to do some additional restructuring before I can fully isolate this `memdata` module. So, I'll leave this as-is.

caput/mpiarray.py
```diff
@@ -251,7 +251,9 @@
 import numpy as np
 import numpy.typing as npt
 
-from caput import fileformats, misc, mpiutil
+from . import mpiutil
+from .memdata import fileformats
+from .memdata._io import open_h5py_mpi
```
ketiltrout, line 256: My take on private modules is: access to them should stop at their parent.  (i.e. they can be accessed by the parent, and by siblings, and by descendants of siblings (nieces?), but not from here, which is what I guess is an "aunt" in this analogy).

I would recommend either renaming `.memdata._io` to `.memdata.io` and just making it public:
```suggestion
from .memdata.io import open_h5py_mpi
```
 or else import the stuff from `.memdata._io` into memdata's `__init__.py`, to hide `.memdata`'s private modules from the outside, and and then here do:
```suggestion
from .memdata import open_h5py_mpi
```

caput/memdata/memh5.py
```diff
@@ -64,7 +64,9 @@
 import h5py
 import numpy as np
 
-from . import fileformats, misc, mpiarray, mpiutil, tools
+from .. import misc, mpiarray, mpiutil, tools
+from . import fileformats
+from ._io import open_h5py_mpi
```
ketiltrout, line 69: (e.g., this access, from a sibling, is fine.)

ketiltrout (approved):
I argue there's a scoping problem with the `memdata._io` private module.

Other than that, this is definitely an improvement.  Moving stuff out of `misc` is always good.

I'm also not in love with `memdata` but I don't have something better.  It's not terrible.  It's good to have a name reminiscent of `memh5` around which this submodule is built (Though I'd argue "mem5h" is also a little imprecisely named these days).  The biggest problem I'd have with the "memdata" name is whether people are going to be surprised to find all the File I/O stuff there, which is the opposite of what I think of when I think of "memory".

Perhaps if `memh5` is rewritten we can rename all of this then?

caput/mpiarray.py
ljgray: Yeah - I'll make these changes for now.

I envision that this I/O stuff will eventually only be used within `memdata`, or whatever it ends up being called, rather than being split between `mpiarray` and `memh5`

PR #302, opened 2025-06-27
Make a new `astro` module

ljgray:
Group a handful of astro-related things into a module.

Also, clean up `units`. Most of this is now imported from `scipy.constants`, with a couple of extra units defined and a special `__getattr__` to handle prefixes.


caput/astro/time.py
```diff
@@ -62,7 +62,7 @@
 
 - As the wrapper is initialised by passing in a ``path=<path>`` option.
 - By setting the environment variable ``CAPUT_SKYFIELD_PATH``
-- If neither of the above is set, the data is place in ``<path to caput>/caput/data/``
+- If neither of the above is set, the data is place in ``<path to caput>/caput/astro/_skyfield_data/``
```
ketiltrout, line 65:
```suggestion
- If neither of the above is set, the data is placed in ``<path to caput>/caput/astro/_skyfield_data/``
```

caput/astro/units.py
```diff
@@ -0,0 +1,50 @@
+"""A set of useful constants and conversions in Astronomy and Cosmology.
```
ketiltrout, line 1: It would be better to call this `caput.astro.constants`.  Its scope has broadened from just being unit definitions.

caput/astro/coordinates/coord.pyx
ketiltrout, line 1: I probably should have caught this earlier, when the `coordinates` submodule was first organised, but this is not good.  We can't have a public module called `coordinates.coord`.  It's too redundant and gives no indication what's in it (compared to the base `coordinates` module).

As everything in this file are related to spherical co-ordinates, I'd suggest turning this to into a private `_spherical.pyx` module and then importing it all into `coordinates.spherical` and have all access happen via that module.  (This is a pretty standard way to do partial Cython modules in a large package)

caput/astro/time.py
```diff
@@ -62,7 +62,7 @@
```
ketiltrout, line 62: The scope of this module is now way too broad to be called `time` anymore.  I'd suggest splitting it into three:

## `caput.astro.skyfield`

Containing the skyfield wrapping stuff (the `SkyfeldClass` and the functions `skyfield_star_from_ra_dec` and `skyfield_time_to_unix`).

## `caput.astro.observer`

Containing the Observer object (which now is used for full ephemerides, and not just time calculations), including the helper function `_fixup_interval_and_step`, which might work better as a private method.

## `caput.astro.time`

Keeping everything else, i.e. all the basic time manipulation functions.

## Also:
* definition of `UT1_S`, `SIDEREAL_S` and `STELLAR_S` should be moved to `astro.constants` so it's all consistent, and then re-imported here.
* Probably we should also import `skyfield_time_to_unix` into `time` and support access from either `caput.astro.skyfield` or `caput.astro.time` for simplicitly.

ketiltrout (changes requested):
On the whole, I think this is a sensible subdivision, but I think there's a need for some further restructuring within the new `caput.astro` module.

caput/astro/coordinates/coord.pyx
ljgray: When I originally made the coordinates module, the idea was to just copy stuff directly from Cora with the intention of eventually reworking it, so I didn't really bother with the internal structure. I guess this is the right time to do that rework

ljgray: Done. Also renamed `sphfunc` to `bessel`, since that's all it contains

ljgray:
Ok, tests are finally working

ketiltrout:
> I think I've addressed all comments re. refactoring. I'm kind of wondering about exporting `Observer` and `SkyfieldWrapper` at the module level -- i.e., `caput.astro.Observer` instead of `caput.astro.observer.Observer`, since those classes are effectively all that's being exported from those modules anyway. Thoughts?

I think that's reasonable, though I'd leave the full path as an option, too (i.e. I wouldn't change, say, `caput.astro.observer` into a private module.)

PR #307, opened 2025-08-18
Add a `pipeline` submodule

ljgray:
Implements the `pipeline` submodule, which groups the pipeline runner, base task infrastructure, and cli into one place.

_There are a lot of files changed, but it's mostly just renaming and imports_

Primary changes:
- Gathers the pipeline runner, pipeline task infrastructure, and pipeline cli into a submodule
- Moves some of the non-essential pipeline infrastructure into `pipeline.extensions`. I think that some of these are no longer relevant, but I'll keep them here for now

This PR got away from me a little bit, with the following unrelated changes:
- Reverted the `darray`/`mpiarray` change. I ended up not really liking this
- Moved the probability distributions in `random` back to `draco`. The core multithreaded random number generator is ok here, but I don't think it made sense to be providing a handful of random distributions
- Gathered some of the fast algorithmic stuff into a submodule. Not quite sure of a name yet.
- Tried to clean up internal dependencies. Specifically, I'm trying to make `astro`, `algs`, `config.py`, etc... dependent only on `util`

There are still a few remaining things to do:
- Split the base `container` infrastructure into pipeline-specific and agnostic. Most of the `container` stuff is agnostic to the pipeline - really, it's just a spec for a certain class of files, and we use it outside of the pipeline all the time. However, there should be a metaclass or something for a `PipelineContainer` that implements anything required by the pipeline (probably just some I/O stuff).
- Remove the residual I/O methods from `mpiarray`. It doesn't make structural sense for that to handle _part_ of the process of writing distributed datasets to `hdf5`.
- Clean up the `SingleTask` base class and remove unneeded task extensions (some of these seems to just be leftover from years ago, but I haven't removed anything yet)
- Fix and improve documentation

I think these changes will be enough to finally merge everything (I mainly want the API to be finalized -- there will probably still be internal updates).


ketiltrout:
Hmm.  What would you say about moving `caput.algs` to `caput.util.fasttools`?

caput/pipeline/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+
+In addition :class:`SingleH5Base`, :class:`IterH5Base`, provide the
+:meth:`read_input` and :meth:`write_output` methods for the most common
+formats.
+
+See the documentation for these base classes for more details.
+
+"""
+
+from . import _core
```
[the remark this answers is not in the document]
ljgray: Nope

ljgray:
> Hmm. What would you say about moving `caput.algs` to `caput.util.fasttools`?

I'm meh on `fasttools`, although it's better than `algs`. I don't really want to move this to `util`, though - since `util` ends up being a dependency for most/all of the other submodules, I don't really want to have compiled code in there.

ketiltrout:
Not wanting it in `util` is reasonable.  Hrm...

ruff.toml
ketiltrout, line 1: Oh: i was wondering, why did you split this out of `pyproject.toml`?

ljgray: I just think that it makes the pyproject file cleaner/less busy. I couldn't find any best practice, but this is what numpy does so I don't think there's anything wrong with it

ketiltrout: Agreed.

ketiltrout:
I'm trying to think about the name from a user point of view.  Do you like any of these:
```
from caput.algs import random
from caput.algos import random
from caput.algorithms import random
from caput.fast import random
from caput.fast_algs import random
from caput.fast_algos import random
from caput.fast_math import random
from caput.math import random
from caput.algs import random
```

ljgray:
@ketiltrout There's one other thing that I'm hoping for thoughts on.

Right now I'm exposing the core pipeline stuff at the `pipeline` module level (i.e., `caput.pipeline.Manager`), whereas tasks are a level down in `caput.pipeline.task`. What are your thoughts on putting the core stuff in `caput.pipeline.core` instead? I'm not sure which approach I prefer.

ljgray:
@ketiltrout I'm mostly done splitting up the cli and runner, so I'd appreciate thoughts. Before I merge this, I also want to pull the jobscript templates out into their own `.yaml` files or something like that, to make that process extensible to other types of job schedulers, but I don't think that will change the current api in any way

ketiltrout:
Just to reiterate: I now think the best option is to just call that module `caput.algorithms`.  My impression is Python modules are only rarely abbreviated, likely because you typically only reference the name once, when you first import it, at which point you can abbreviate it yourself, if you want.  e.g.
```
import matplotlib.pyplot as plt
```
or, even,
```
import caput.algorithms as algs
```

Issue #303, opened 2025-07-05
Re-investigate parallel IO compression in `memh5`


ljgray:
Previously, compression was disabled when using parallel IO for hdf5 files from `memh5`. Discussion (https://github.com/chime-experiment/Pipeline/issues/33) and (https://forum.hdfgroup.org/t/crash-when-writing-parallel-compressed-chunks/6186/21).

By removing [this chunk](https://github.com/radiocosmology/caput/blob/7625b5ecd6d5787c67b8be4a76b31f2b8adc6097/caput/memh5.py#L2691) from `memh5`, I was able to successfully write a smallish file using parallel IO with compression/chunking enabled. I can't say for sure if this means that the issue is actually resolved, but a few years and several versions of hdf5 later, we should probably investigate this issue again.

ljgray:
@jmaceachern Tagging you as an fyi

ljgray:
Ok, it seems like there were a bunch of parallel compression fixes and improvements in HDF5 1.13.1 (an experimental release) (https://www.hdfgroup.org/2022/03/04/parallel-compression-improvements-in-hdf5-1-13-1/) which are now integrated into HDF5 1.14.x. Combined with the testing I've done, I think we're ok to re-enable this, so I've made a PR to do that.

PR #308, opened 2025-08-27
Re-enable parallel compression for hdf5 files

ljgray:
As discussed in #303, it looks like the parallel compression issues with HDF5 are resolved since 1.13.1. 

This PR removes the code that forcibly disabled parallel compression, so writing parallel HDF5 should now behave the same way as writing parallel `zarr`. It also fixes a test that assumed that compression would always be disabled for hdf5 files.

### DIFF
--- a/caput/fileformats.py
+++ b/caput/fileformats.py
@@ -10,8 +10,8 @@ logger = logging.getLogger(__name__)
 
 try:
     import zarr
-except ImportError as err:
-    logger.info(f"zarr support disabled. Install zarr to change this: {err}")
+except ImportError:
+    logger.debug("zarr support disabled. Install zarr to change this.")
     zarr_available = False
 else:
     zarr_available = True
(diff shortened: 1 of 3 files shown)


caput/fileformats.py
```diff
@@ -72,11 +72,6 @@ class HDF5(FileFormat):
 
     module = h5py
 
-    @staticmethod
-    def compression_enabled():
-        """Disable compression and chunking due to bug: https://github.com/chime-experiment/Pipeline/issues/33."""
-        return False
-
```
ketiltrout, line 79: How certain are you that we won't want disable compression in the future?  A less invasive change that accomplishes the same thing would be to simply return `True` here (and then not delete the call to this function in mem5h).

It could also permit disabling compression on a case-by-case basis by changing the value returned from this method.  Would that be useful?

caput/memh5.py
```diff
@@ -2685,16 +2685,6 @@ def _prepare_compression_args(dset):
             )
         compression_kwargs["chunks"] = getattr(dset, "chunks", None)
 
-        # disable compression if not enabled for HDF5 files
-        # https://github.com/chime-experiment/Pipeline/issues/33
-        if (
-            to_file
-            and file_format == fileformats.HDF5
-            and not fileformats.HDF5.compression_enabled()
```
ketiltrout, line 2693: See comment above about the `compression_enabled` method.

caput/fileformats.py
ketiltrout: Do we have a way to enforce minimum HDF5 library version?

If not, an even slicker option would be to check the library version and return True/False based on the needed compression support.

ketiltrout (commented):
I think it would be at least worth thinking about what to do in the case where the HDF5 library used by caput isn't new enough for this to work.  AFAIK, the HDF5 CAPI verison is retrievable from `h5py`.

I don't see a way of tiying `h5py` verison to HDF5lib version, but maybe checking it at runtime would be enough.

ljgray:
Unfortunately I don't know firsthand, but from the discussion in the hdf forum thread I think there was some combination of hanging and crashing, depending on the specific combination of compression/chunksize. So, I would lean towards having a more informative error message

PR #311, opened 2025-10-28
Update docs

ljgray:
This PR updates docstrings and adds typing throughout the module. The project layout is unchanged, as is the functionality (outside of a couple of small name changes in private files).

It also gets the sphinx API reference working - I still need to make some updates, but half the submodules weren't previously being included.

## Typing
I've implemented type hints everywhere, along with some custom types in `mpiarray` and `memh5`. I think I've mostly followed standard practice (insofar as it exists), but it should be revisited later (especially with custom types, which have much better support in python 3.12+). 

## Docs
~I've added/updated docstrings everywhere, and got the readthedocs API reference to (mostly) work properly (i.e., docs are created for everything). I added the `sphinx_autodoc_typehints` extension, which mostly seems to work, but it still has issues with overloads and decorators. I'm going to investigate using [sphinx-autoapi](https://sphinx-autoapi.readthedocs.io/en/latest/) instead of autodoc, as autoapi doesn't actually have to install and import the project, but that's outside the scope of this PR.~
I ended up switching to [sphinx-autoapi](https://sphinx-autoapi.readthedocs.io/en/latest/) -- I just kept running into all sorts of errors with `autodoc` and `autosummary`. It seems to work pretty well, and it's significantly faster than `autodoc`. However, some extensions the rely on importing code seem to run into issues (ex: `viewcode` doesn't work). I ended up switching to `linkcode` to generate links to github, which also required some hacking because the module info that this extensions receives is a bit different from what it expects. I _think_ that I've got everything working now, though.

For whatever reason, the theme that I'm using isn't compatible with `numpydoc`, but it works with `napoleon`. I think that this is fine, but I'm running into a few weird things that I still need to debug.

Finally, it seems that `numpydoc` and `napoleon` don't currently work very well with type hints, as noted with the duplicate `Return Type` sections. I'm going to experiment with this a little bit -- from what I can tell, the current recommendation is to include types in the standard docstrings and implement type hints in stub files (`.pyi`). I think that there just isn't a globally accepted way of handling this in python (which seems to be the general trend with anything typing-related).

## Tests
I had to make some adjustments to the tests to get them to actually run:
- The API reference needs to run in editable mode. I still haven't quite figured out why, but it seems to be related to the cython code.
- Including the `--doctest-modules` flag results in import errors for cython files that I haven't been able to work around, again only with standard installs. At the moment, the tests pass fine if I run the doctest separately with `--doctest-ignore-import-errors`. I'd like to figure out what's going on, but I think this has more to do with the way that pytest is configured as opposed to issues with our code, so I'm going to leave it for a future PR where I rework the tests. This also seems to mostly be related to cython code.
- `numcodecs` doesn't have wheels available for python 3.14 yet, and it doesn't build properly on macos. For now, I'm just skipping that test, but I'll revisit it.

## Other
I switched from using `black` to `ruff format`. There's no difference in the rules (`ruff format` explicitly tries to be a drop-in replacement), but it means we only need one tool (`ruff`).

## Notes
I intend to follow up this PR with a couple of final, small API changes - I vaguely want to decide exactly what the `memdata` API should look like before I start on backend updates. The final API changes should be sufficient to finally merge the `refactor-dev` branch along with updates in other repos, since any future changes should keep the same API.

I'm also going to try to write actual docs and modernize the test suite, but this can happen after the API is finalized and merged into `master`/`main`.

### DIFF
--- a/caput/pipeline/tasklib/flow.py
+++ b/caput/pipeline/tasklib/flow.py
@@ -74,6 +74,29 @@ class AccumulateList(MPILoggedTask):
         return items if self.group_size is None else None
 
 
+class GatherInputs(MPILoggedTask):
+    """Gather an arbitrary number of inputs into a list.
+
+    The relevant number of inputs is determined by the task
+    configuration keys.
+    """
+
+    def next(self, *inputs):  # noqa: D417
+        r"""Gather the inputs into a list.
+
+        Parameters
+        ----------
+        \*inputs : list[Any]
+            Arbitrary number of inputs.
+
+        Returns
+        -------
+        output : list[Any]
+            A list containing all the inputs.
+        """
+        return list(inputs)
+
+
 class Delete(MPILoggedTask):
     """Delete pipeline products to free memory."""
 


ketiltrout:
Do you have the rendered docs up somewhere?

ljgray:
> Also, you seem to have made the same mistake I made in alpenhorn: According to the numpy doc standard, the "Returns" section should be formatted like:
> 
> ```
> Returns
> -------
> return_type:
>     Description of returned value.
> ```
> 
> I had, instead of `return_type`, put the name of the returned value in the first line. (Numpydoc doesn't seem to support naming the returned value). This explains things like: https://caput--311.org.readthedocs.build/en/311/_autosummary/_autosummary/caput.mpiarray.html#caput.mpiarray.MPIArray.global_shape where the return _type_ is stated to be "global_shape", which isn't a type.

Interesting - do you know how this is supposed to be handled when the return type is provided through type hints? I wonder if that's why there are  duplicate return values in some places.

ljgray:
> I'll go through this more thoroughly, but I'm seeing some strange things like this:
> 
> https://caput--311.org.readthedocs.build/en/311/_autosummary/_autosummary/caput.mpiarray.html#caput.mpiarray.MPIArray
> 
> where the parameter list is given twice. I think this is because it's combining the class-level docstring with the docstring from the `__new__` method. Not sure how to best deal with that.

Good catch. Luckily this shouldn't affect too much of the code, but I'm sure there are guideline somewhere.

ketiltrout:
Ah, you seem to be using that already...

ketiltrout:
Great.  I'll take a closer look.

caput/algorithms/fft/fftw.py
```diff
@@ -1,16 +1,16 @@
...
 
 Only forward and reverse complex->complex transforms
 are currently supported.
 
 Examples
 --------
-The core of this module is the :class:`FFT`, which essentially just
-abstracts the :class:`pyfftw:FFTW` in the simplest way.
+The core of this module is the :py:class:`FFT`, which essentially just
+abstracts the :py:class:`pyfftw:FFTW` in a simple way.
```
ketiltrout:
```suggestion
abstracts the :py:class:`pyfftw:FFTW` class in a simple way.
```

caput/algorithms/fft/fftw.py
```diff
@@ -30,80 +30,67 @@
...
 class FFT:
-    """Faster FFTs with FFTW."""
+    """Create FFTW objects for repeat use.
+
+    This implementation is most efficient when used to repeatedly
+    apply ffts to arrays with the same shape and dtype, because a
+    single, highly optimised pathway can be used with a single
+    initialisation.
+
+    Even for a single use this will typically
```
ketiltrout:
```suggestion
    Even for a single use, this will typically
```

ljgray:
I'm also completely fine with changing the colour scheme of the docs, I just picked something that seemed unique...

ketiltrout:
Hmm.  Also, I guess: I don't know if it's just a result of the test-version made for this PR, but the multiple levels of `_autosummary` in the URLs is odd.

caput/algorithms/invert_no_zero/__init__.py
```diff
@@ -12,24 +17,26 @@
 ]
 
 
-def invert_no_zero(x, out=None):
+def invert_no_zero(
+    x: npt.ArrayLike[np.number], out: npt.ArrayLike[np.number] | None = None
+) -> npt.ArrayLike[np.number]:
     """Return the reciprocal, but ignoring zeros.
 
     Where `x != 0` return 1/x, or just return 0. Importantly this routine does
```
ketiltrout:
```suggestion
    Where ``x != 0`` return ``1/x``; where ``x == 0``, return 0. Importantly this routine does
```

I think this needs to be more explicit.

caput/algorithms/invert_no_zero/__init__.py
```diff
@@ -12,24 +17,26 @@
...
 
     Parameters
     ----------
-    x : np.ndarray
+    x
         Array to invert
-    out : np.ndarray, optional
-        Output array to insert results
+    out
+        Output array to insert results. Default is None,
```
ketiltrout:
```suggestion
        Output array where the result is stored. Default is None,
```

ketiltrout (commented):
This is going to take me too long if I wait to go through it all in one go, so here's some copy edits for the `algorithms` submodule

ljgray:
> > Yeah, I think it's because of the nested directory structure but it would be nice if `_autosummary` didn't appear in the URL in the first place. I'll see if there's a way to configure that
> 
> Maybe it has something to do with the seemingly unnecessary "Submodules" sections in the ToC?

I'm mildly confused by this, because I was under the impression that autodoc should be able to handle nested directories, but if I remove the `Submodules` section from an `__init__` file, nothing in the API reference ends up getting generated.

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+    "thetaphi_plane_cart",
+]
+
+#
+def sph_to_cart(
+    sph_coords: npt.NDArray[np.integer | np.floating],
+) -> np.ndarray[np.float64]:
+    """Convert a vector in Spherical Polar coordinates to Cartesians.
+
+    This routine is OpenMP parallel.
```
ketiltrout:
```suggestion
    """Convert a vector in spherical polar co-ordinates to Cartesian co-ordinates.

    This routine is OpenMP-parallel.
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+
+    Parameters
+    ----------
+    sph_coords
+        A vector (or array of) in spherical polar co-ordinates. Values should be
```
ketiltrout:
```suggestion
        A vector (or array) of spherical polar co-ordinates. Values should be
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+    Parameters
+    ----------
+    arr1, arr2
+        Two arrays of vectors in spherical polars [theta, phi], (or
+        alternatively as [theta, phi]). Should be broadcastable against each
+        other.
+
+    Returns
+    -------
+    An array of the dotted vectors.
```
ketiltrout:
```suggestion
    An array of the scalar products.
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+
+    Parameters
+    ----------
+    sph_arr
+        Angular positions (in spherical polar co-ordinates).
+
+    Returns
+    -------
+    Unit vectors in the theta and phi directions (still in spherical
+    polars).
```
ketiltrout:
```suggestion
    Theta and phi directions (still in spherical polar co-ordinates) for the unit vectors.
```

The point here (perhaps poorly executed) is to emphasise that this function returns a 2-tuple of arrays of angles, not a single array of angle pairs.

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+    sph_coords: npt.NDArray[np.integer | np.floating],
+) -> np.ndarray[np.float64]:
+    """For each position, return theta, phi tangent vectors in cartesian coordinates.
+
+    This routine is OpenMP parallel.
+
+    Parameters
+    ----------
+    sph_coords
+        Angular positions (in spherical polar co-ordinates).
```
ketiltrout:
```suggestion
        Angular unit vectors (in spherical polar co-ordinates) packed as [r, theta, phi] along the last axis.
```
My reading of the code here is that these are unit vectors, not full positions.

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+
+    Returns
+    -------
+    Unit vectors in the theta and phi directions now in cartesian coordinates.
+    """
+    ...
+
+#
+def norm_vec2(vec2: npt.NDArray[np.floating]) -> None:
+    """For an array of 2D vectors, normalise each to unit length *inplace*.
```
ketiltrout:
```suggestion
    """For an array of 2D vectors, normalise each to unit length *in-place*.
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+    npoints: int | np.integer,
+) -> np.ndarray[np.floating]:
+    """Return the intermediate points on the great circle between points `sph1` and `sph2`.
+
+    Parameters
+    ----------
+    sph1, sph2
+        Points on sphere, packed as [theta, phi]
+    npoints
+        Number of intermediate points
```
ketiltrout:
```suggestion
        Number of intermediate points to compute.
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+    """
+    ...
+
+#
+def cosine_rule(
+    mu: npt.ArrayLike[np.float64],
+    x1: npt.ArrayLike[np.float64],
+    x2: npt.ArrayLike[np.float64],
+) -> np.ndarray[np.float64]:
+    """Calculate the distance between a grid of points.
```
ketiltrout:
```suggestion
    """Calculate the distances between a grid of points.
```

caput/astro/coordinates/_spherical.pyi
```diff
@@ -0,0 +1,167 @@
...
+
+    This is a somewhat niche implementation intended to help calculate multi-distance
+    angular power spectra.
+
+    Parameters
+    ----------
+    mu
+        Angular separation in cos(theta)
```
ketiltrout:
```suggestion
        Angular separations in ``cos(theta)``
```

caput/astro/coordinates/spherical.py
```diff
@@ -1,81 +1,71 @@
...
 def sph_to_ground(ha, lat, dec):
     """Get the ground based XYZ coordinates.
 
     All input angles are radians. HA, DEC should be in CIRS coordinates.
 
     Parameters
     ----------
-    ha : array_like
+    ha
         The Hour Angle of the source to fringestop too.
```
ketiltrout:
```suggestion
        The CIRS hour angle of the source in radians.
```

caput/astro/coordinates/spherical.py
```diff
@@ -1,81 +1,71 @@
...
         The latitude of the observatory.
-    dec : array_like
+    dec
         The declination of the source.
 
     Returns
     -------
-    x, y, z : array_like
-        The projected angular position in ground fixed XYZ coordinates.
+    The projected angular position in ground fixed XYZ coordinates.
```
ketiltrout:
```suggestion
    The projected angular position in ground-fixed XYZ coordinates.
```

caput/astro/coordinates/spherical.py
```diff
@@ -115,28 +132,45 @@ def ground_to_sph(x, y, lat):
...
+    z: np.number | npt.NDArray[np.number] = 0.0,
+) -> np.number | np.ndarray[np.number]: ...
 def projected_distance(ha, lat, dec, x, y, z=0.0):
     """Return the distance project in the direction of a source.
 
     Parameters
     ----------
-    ha : array_like
+    ha
         The Hour Angle of the source to fringestop too.
```
ketiltrout:
```suggestion
        The CIRS hour angle of the source in radians.
```

caput/astro/coordinates/spherical.py
```diff
@@ -115,28 +132,45 @@ def ground_to_sph(x, y, lat):
...
-    lat : array_like
+    lat
         The latitude of the observatory.
-    dec : array_like
+    dec
         The declination of the source.
-    x : array_like
+    x
         The EW coordinate in wavelengths (increases to the E)
```
ketiltrout:
```suggestion
        The East-West coordinate in wavelengths (increases to the East)
```

caput/astro/coordinates/spherical.py
```diff
@@ -115,28 +132,45 @@ def ground_to_sph(x, y, lat):
...
-    y : array_like
+    y
         The NS coordinate in wavelengths (increases to the N)
```
ketiltrout:
```suggestion
        The North-South coordinate in wavelengths (increases to the North)
```

caput/astro/constants.py
```diff
@@ -1,36 +1,58 @@
...
+"""Solar mass in kg."""
+
+second: float = 1.0
+"""One second in seconds."""
+
+t_sidereal: float = 23.9344696 * hour
+"""One sidereal day in seconds, equal to `23.9344696 * hour`."""
+
+a_rad: float = 4.0 * Stefan_Boltzmann / c
+"""Radiation constant (in J m^{-3} K^{-4}), equal to `4 * Stefan_Boltzmann / c`."""
```
ketiltrout:
```suggestion
"""Radiation constant (in J m\ :sup:`-3` K\ :sup:`-4`), equal to `4 * Stefan_Boltzmann / c`."""
```
(I think this is the correct ReST syntax for superscripts.)

caput/astro/observer.py
```diff
@@ -3,7 +3,7 @@
 Local Time Utilities
 ====================
 
-Routines which are location specific are grouped into the location aware class :py:class:`Observer`.
+Routines are provided through the location aware class :py:class:`Observer`.
```
ketiltrout:
```suggestion
Routines are provided through the location-aware class :py:class:`Observer`.
```

caput/astro/observer.py
```diff
@@ -231,35 +174,46 @@ def unix_to_lsa(self, time):
...
+        self, lsa: npt.ArrayLike[np.floating], time0: npt.ArrayLike[np.floating]
+    ) -> np.ndarray[np.floating]: ...
     def lsa_to_unix(self, lsa, time0):
         """Convert a Local Stellar Angle (LSA) on a given day to a UNIX time.
 
         Parameters
         ----------
-        lsa : scalar or np.ndarray
+        lsa
             Local Earth Rotation Angle degrees to convert.
```
ketiltrout:
```suggestion
            Local Stellar Angle to convert, in degrees.
```
We've explained the relationship between LSA and ERA above.

caput/astro/observer.py
```diff
@@ -231,35 +174,46 @@ def unix_to_lsa(self, time):
...
-        time0 : scalar or np.ndarray
+        time0
             An earlier time within 24 sidereal hours. For example,
```
ketiltrout:
```suggestion
            An earlier UNIX time within 24 sidereal hours. For example,
```
Just to be clear.

caput/astro/skyfield.py
```diff
@@ -1,40 +1,25 @@
...
-==================
-
-- :py:class:`SkyfieldWrapper`
-
-This module provides an interface to Skyfield which stores the required datasets
+This module provides an interface to `skyfield`_ which stores the required datasets
 (timescale data and an ephemeris) in a fixed location. The location is
 determined by the following (in order):
 
 - As the wrapper is initialised by passing in a ``path=<path>`` option.
```
ketiltrout:
```suggestion
- By passing in a ``path=<path>`` option as the wrapper is initialised.
```

caput/astro/time.py
```diff
@@ -169,6 +167,10 @@ def era_to_unix(era, time0):
     return time0 + diff_time - leap_seconds
 
 
+@overload
+def unix_to_datetime(unix_time: np.floating) -> datetime: ...
+@overload
+def unix_to_datetime(unix_time: npt.ArrayLike[np.floating]) -> np.ndarray[datetime]: ...
 @arraytools.vectorize(otypes=[object])
 def unix_to_datetime(unix_time):
     """Converts unix time to a :class:`~datetime.datetime` object.
```
ketiltrout:
```suggestion
    """Converts UNIX time to a :class:`~datetime.datetime` object.
```

caput/astro/time.py
```diff
@@ -177,23 +179,26 @@ def unix_to_datetime(unix_time):
 
     Parameters
     ----------
-    unix_time : float
+    unix_time
         Unix/POSIX time.
```
ketiltrout:
```suggestion
        UNIX time.
```

caput/containers/__init__.py
```diff
@@ -1,13 +1,62 @@
...
+:py:class:`.TableBase`.
 
-Submodules
-----------
-.. autosummary::
-   :toctree: _autosummary
+Examples
+--------
+Containers are defined by subclassing :py:class:`ContainerBase` and specifying
+the axes and dataset specifications:
```
ketiltrout:
```suggestion
the axes and datasets:
```
To avoid saying "specify/specification" twice.

caput/containers/__init__.py
```diff
@@ -1,13 +1,62 @@
...
+...                 "freq",
+...             ],
+...             "dtype": np.float32,
+...             "initialise": True,
+...             "compression": None,
+...             "distributed": False,
+...         },
+...     }
+
+When a containers is instantiated, the datasets defined in :py:attr:`_dataset_spec` are
```
ketiltrout:
```suggestion
When a container is instantiated, the datasets defined in :py:attr:`_dataset_spec` are
```

caput/containers/__init__.py
```diff
@@ -1,13 +1,62 @@
-"""caput.containers.
+"""A high-level in-memory data container format for :py:mod:`caput.pipeline`.
 
-A high-level in-memory data container format for `caput.pipeline`.
+Containers are built on top of :py:class:`~caput.memdata.BasicCont` and provide a structured way
```
ketiltrout: Is there a way to get a mention of memdata here.  This renders as

> Containers are built on top of [BasicCont](https://caput--311.org.readthedocs.build/en/311/autoapi/caput/memdata/BasicCont.html#caput.memdata.BasicCont) and provide

The link works fine, but someone who hasn't learned about memdata yet might think `BasicCont` is part of `containers` (given its name).

Maybe:

```suggestion
Containers are built on top of memdata's :py:class:`~caput.memdata.BasicCont` and provide a structured way
```
with a deliberately unlinked "memdata"?

caput/containers/__init__.py
ketiltrout:
```suggestion
Containers are subclasses of :py:class:`ContainerBase`.  New containers are defined by specifying
the axes and datasets when subclassing:
```
The worry here being people might think that _only_ `ContainerBase` should be subclassed (instead of, say, subclassing `TableBase`).

Also remove a duplication of "specify/specification".

caput/containers/__init__.py
ketiltrout:
```suggestion
When a container is instantiated, the datasets defined in :py:attr:`_dataset_spec` are
```

caput/containers/__init__.py
```diff
@@ -1,13 +1,62 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+Axes
+~~~~
+A container must specify a set of named axes that define the dimensions of its
+datasets. Each axis must be associated with a corresponding entry in the `index_map`
+dataset. Axes are specified via the :py:attr:`_axes` class attribute, which is a list
+of strings, and are inherited from parent classes.
+
+Dataset Spec
+~~~~~~~~~~~~
```
ketiltrout:
```suggestion
^^^^^^^^^^^^
```

caput/containers/_basic.py
```diff
@@ -9,28 +18,28 @@ class DataWeightContainer(ContainerBase):
 
     This is meant such that tasks can operate generically over containers with this
```
ketiltrout:
```suggestion
    This is meant to be a general-purpose container such that tasks can operate generically over containers with this
```

caput/containers/_core.py
```diff
@@ -1,75 +1,86 @@
...
+    Optional Parameters
+    ~~~~~~~~~~~~~~~~~~~
+    Some combination of the following parameters must br provided when creating
+    a new container. In particular, the container requires axis definitions to
+    be provided for all axes in :py:attr:`~.ContainerBase._axes`, whether by
+    direct keyword arguments, or by copying from another container.
+
+    data_group : :py:class:`~caput.memdata.MemDiskGroup`
         A container to pass through for making a shallow copy. This is used by
         routine like `caput.tod.concatenate` and generally shouldn't be used
```
ketiltrout:
```suggestion
        routines like `caput.tod.concatenate` and generally shouldn't be used
```

caput/containers/_core.py
```diff
@@ -1,75 +1,86 @@
...
         A container to copy datasets from. Any dataset which an axis whose definition
         has been explicitly set (i.e. does not come from `axes_from`) will not be
         copied.
-    copy_from : `memh5.BasicCont`, optional
-        Set `axes_from`, `attrs_from` and `dsets_from` to this instance if they are
+    copy_from : :py:class:`~caput.memdata.BasicCont`, optional
+        Set ``axes_from``, ``attrs_from`` and ``dsets_from`` to this instance if they are
         not set explicitly.
     skip_datasets : bool, optional
         Skip creating datasets. They must all be added manually with
```
ketiltrout:
```suggestion
        Skip creating datasets. Instead, they will all need to be added manually with
```

caput/containers/_core.py
```diff
@@ -314,26 +326,25 @@ def _ensure_chunked(self):
                 ]["compression_opts"]
 
     @property
-    def datasets(self):
+    def datasets(self) -> ro_dict[str, memdata.MemDataset]:
         """Return the datasets in this container.
```
ketiltrout:
```suggestion
        """A read-only view of the datasets in this container.
```

caput/containers/_util.py
```diff
@@ -1,28 +1,38 @@
...
 
-from ..memdata import memh5
+from ..memdata import _memh5
 from ..mpiarray import _apply_sel
 from ._core import ContainerBase
 
 
-def empty_like(obj, **kwargs):
-    """Create an empty container like `obj`.
+def empty_like(obj: ContainerBase, **kwargs: dict) -> ContainerBase:  # noqa: D417
```
ketiltrout: I think this could benefit from a parameter name less-generic than `obj`, but I don't have a good suggestion.

Maybe `ref`, for "reference"?

caput/containers/tod.py
```diff
@@ -1,39 +1,59 @@
...
 
-class TOData(memh5.BasicCont):
+__all__ = ["TODReader", "TOData", "concatenate"]
+
+
+class TOData(memdata.BasicCont):
     """Time ordered data.
 
-    Inherits from :class:`caput.memh5.BasicCont`. A data container in with all
+    Inherits from :py:class:`~caput.memdata.BasicCont`. A data container in with all
```
ketiltrout:
```suggestion
    Inherits from :py:class:`~caput.memdata.BasicCont`. A data container with all
```

caput/containers/tod.py
```diff
@@ -1,39 +1,59 @@
...
     the functionality of its base class but with the concept of a time axis
     which can be concatenated over. Currently the time axis must be the fastest
     varying axis is present.
```
ketiltrout:
```suggestion
    which can be concatenated over. Currently the time axis must be the
    fastest-varying axis.
```

caput/containers/tod.py
```diff
@@ -50,53 +70,55 @@ def time(self):
...
-        Accepts any parameter for :func:`concatenate` (which controls the
+        files: FileLike | Sequence[FileLike],
+        data_group: GroupLike | None = None,
+        start: int | dict | None = None,
+        stop: int | dict | None = None,
+        datasets: Sequence[str] | None = None,
+        dataset_filter: Callable | None = None,
+        **kwargs: dict,
+    ) -> TOData:
+        r"""Create new data object by concatenating a series of objects.
```
ketiltrout:
```suggestion
        r"""Create new data object by concatenating a series of files.
```

caput/containers/tod.py
```diff
@@ -50,53 +70,55 @@ def time(self):
...
+
+        Accepts any parameter for :py:func:`concatenate` (which controls the
```
ketiltrout:
```suggestion
        Accepts any parameter supported by :py:func:`concatenate` (which controls the
```
Also: the ":py:func:`concatenate`" isn't being turned into a link for some reason.

caput/containers/tod.py
```diff
@@ -50,53 +70,55 @@ def time(self):
...
+        dataset_filter : Callable, optional
             Function for preprocessing all datasets.  Useful for changing data
             types etc. Takes a dataset as an argument and should return a
-            dataset (either h5py or memh5). Optionally may accept a second
+            dataset (either h5py or memdata). Optionally may accept a second
             argument that is slice along the time axis, which the filter should
             apply.
-        **kwargs : dict
+        \**kwargs : dict
             Arbitrary keyword arguments.
```
ketiltrout:
```suggestion
            Other keyword arguments are passed on to the class's `from_file` method.
```

caput/containers/tod.py
```diff
@@ -125,26 +147,30 @@ def convert_time(time):
         return time
 
 
-class Reader:
-    """Provides high level reading of time ordered data.
+class TODReader:
+    r"""Provides high level reading of time ordered data.
```
ketiltrout:
```suggestion
    r"""Provides high-level reading of time ordered data.
```

caput/containers/tod.py
```diff
@@ -166,83 +192,84 @@ def __init__(self, files, file_format=None):
...
     @property
-    def datasets(self):
+    def datasets(self) -> list[str]:
         """Datasets available in data files."""
         return self._datasets
 
     @property
-    def time_sel(self):
+    def time_sel(self) -> tuple[int, int]:
         """Start and stop indices to read in the frequency axis.
```
ketiltrout:
```suggestion
        """Start and stop indices to read in the time axis.
```

ketiltrout (commented):
For the `containers` module

caput/containers/__init__.py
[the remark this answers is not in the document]
ketiltrout: duplicated below.

caput/containers/__init__.py
```diff
@@ -1,13 +1,62 @@
...
+New containers should generally inherit from :py:class:`.ContainerBase` or
+:py:class:`.TableBase`.
 
-Submodules
-----------
-.. autosummary::
-   :toctree: _autosummary
+Examples
+--------
+Containers are defined by subclassing :py:class:`ContainerBase` and specifying
```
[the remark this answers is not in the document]
ketiltrout: duplicated below

caput/memdata/__init__.py
```diff
@@ -1,24 +1,41 @@
-"""caput.memdata.
+"""A backend format for in-memory or on-disk hdf5-like datasets.
 
-A backend format for in-memory or on-disk hdf5-like datasets.
+It is useful to have a consistent API for data that is independent
+of whether that data lives on disk or in memory. :py:mod:`h5py` provides this to a
+certain extent, having :py:class:`h5py.Dataset` objects that act very much like
+:py:mod:`numpy` arrays. :py:mod:`~caput.memdata` extends this, providing an in-memory
```
ketiltrout:
```suggestion
:py:mod:`numpy` arrays. :py:mod:`~caput.memdata` extends this, providing in-memory
```

caput/memdata/__init__.py
```diff
@@ -1,24 +1,41 @@
...
+containers, analogous to :py:class:`h5py.Group`, :py:class:`h5py.AttributeManager` and
+:py:class:`h5py.Dataset` objects.
 
-Submodules
-----------
-.. autosummary::
-   :toctree: _autosummary
+In addition to these basic classes that copy the :py:mod:`h5py` API, a higher
+level data container is provided that utilizes these classes along with the
```
ketiltrout:
```suggestion
In addition to these basic classes that copy the :py:mod:`h5py` API, a higher-level
data container is provided that utilizes these classes along with the
```

caput/memdata/_io.py
```diff
@@ -1,26 +1,14 @@
...
-    use_mpi: bool = True,
-    comm: Optional["MPI.Comm"] = None,
-) -> h5py.File: ...
-@overload
-def open_h5py_mpi(
-    f: h5py.Group, mode: str, use_mpi: bool = True, comm: Optional["MPI.Comm"] = None
-) -> h5py.Group: ...
+
+
 def open_h5py_mpi(f, mode, use_mpi=True, comm=None):
```
ketiltrout: This doesn't show up in the docs.  Should it?

caput/memdata/_io.py
```diff
@@ -84,29 +72,23 @@ class lock_file:
         Final name for the file.
     preserve : bool, optional
         Keep the temporary file in the event of failure.
-    comm : MPI.COMM, optional
+    comm : MPI.Comm | None, optional
         If present only rank=0 will create/remove the lock file and move the
```
ketiltrout:
```suggestion
        If not `None`, only rank=0 will create/remove the lock file and move the
```

ketiltrout (commented):
I started looking at `memdata`, but I think a lot of the intended docs aren't actually being rendered.

Maybe I'll wait on this part.

ljgray:
> > > I started looking at `memdata`, but I think a lot of the intended docs aren't actually being rendered.
> > > Maybe I'll wait on this part.
> > 
> > 
> > Hmm. I'm thinking that private modules aren't being rendered, so anything that isn't re-exported in `__init__` gets skipped. That should be easy to resolve...
> 
> Actually, this ends up being really annoying because it generates duplicate entries for everything that's re-exported...

Ok, this seems to be an inherent limitation of `sphinx-autodoc` - even if we generate references for all private members, it still fails to resolve the inheritance tree for public classes exposed in __init__ that inherit from private classes in a different private file.

caput/memdata/fileformats.py
```diff
@@ -1,172 +1,313 @@
...
         """
         raise NotImplementedError
 
-    @staticmethod
-    def compression_kwargs(compression=None, compression_opts=None, compressor=None):
+    @classmethod
+    def compression_kwargs(
+        cls, compression=None, compression_opts=None, compressor=None
+    ):
         """Sort compression arguments in a format expected by file format module.
```
ketiltrout: "Sorting" usually implies changing the order of data, and not its form, but it's not clear that this is what's implied here, since the word "format" is being used. 

I'd perhaps say:

```suggestion
        """Sort compression arguments into an order expected by the file format module.
```

to reinforce that sorting is what's happening, or else:


```suggestion
        """Arrange compression arguments into a format expected by the file format module.
```

if the operation is more general than simple sorting.

caput/memdata/fileformats.py
```diff
@@ -1,172 +1,313 @@
...
+        Not implemented in base class.
+
         Parameters
         ----------
-        compression : str or int
+        compression : str | int | None
             Name or identifier of HDF5 compression filter.
-        compression_opts
+        compression_opts : tuple | None
             See HDF5 documentation for compression filters.
```
ketiltrout:
```suggestion
            Options for the selected filter.  See HDF5 documentation for compression filters.
```

caput/memdata/fileformats.py
```diff
@@ -1,172 +1,313 @@
...
-        compressor : `numcodecs` compressor
-            As required by `zarr`.
+        compressor : :py:mod:`numcodecs` compressor
+            As required by :py:mod:`zarr`.
```
ketiltrout:
```suggestion
            The compressor to use.
```
If we want to make it more generic.

caput/memdata/fileformats.py
```diff
@@ -257,15 +400,16 @@ def check_file_format(filename, file_format, data):
...
+    data : Any
+        If this is an :py:class:`h5py.Group` or :py:class:`zarr.Group`, it will be used to
+        guess or confirm the file format.
 
     Returns
     -------
-    file_format : HDF5 or Zarr
-        File format.
+    file_format : FileFormat
+        :py:class:`FileFormat` instance.
```
ketiltrout:
```suggestion
        The detected :py:class:`FileFormat`.
        
    Raises
    ------
    ValueError:
       The detected file format did not match `file_format`.
```

caput/memdata/fileformats.py
ketiltrout:
```suggestion
    """Attempt to detect the file format of a file.
```

ketiltrout:
Hmm, that's all in one file. Did GitHub forget about comments I had for the other files?

ketiltrout:
> Hmm, that's all in one file. Did GitHub forget about comments I had for the other files?

No, I guess not.  That looks like everything I had for `memdata`.

caput/pipeline/extensions/__init__.py
```diff
@@ -1,4 +1,10 @@
-"""Extensions for `TaskBase` and helper code to run pure python pipelines."""
+"""Extensions for :py:class:`~caput.pipeline.Task` and helper code to run pure python pipelines.
```
ketiltrout:
```suggestion
"""Extensions for :py:class:`~caput.pipeline.Task` and helper code to run pure Python pipelines.
```

caput/pipeline/extensions/_configtypes.py
```diff
@@ -1,38 +1,40 @@
...
     Parameters
     ----------
-    default
+    default : str | FileFormat | None, optional
         A string or type object specifying the fileformat
 
     Returns
     -------
-    prop
+    file_format : FileFormat
```
ketiltrout:
```suggestion
    prop : Property
```
This _function_ returns a `Property`.  The `Property` is the thing that'll return a `FileFormat`.

caput/pipeline/extensions/_configtypes.py
```diff
@@ -1,38 +1,40 @@
...
 
 __all__ = ["file_format"]
 
 
-def file_format(default: str | fileformats.FileFormat | None = None) -> Property:
+def file_format(default=None):
     """Property type that accepts only "zarr", or "hdf5".
 
-    Returns the selected `caput.fileformat.FileFormat` subclass or `caput.fileformats.HDF5` if `value == default`.
+    Returns the selected :py:class:`~caput.fileformats.FileFormat` subclass or
```
ketiltrout:
```suggestion
    The created Property returns the selected :py:class:`~caput.fileformats.FileFormat` subclass or
```
This function returns a `Property`

caput/pipeline/extensions/_configtypes.pyi
```diff
@@ -0,0 +1,6 @@
+from ...config import Property
+from ...memdata import fileformats
+
+__all__ = ["file_format"]
+
+def file_format(default: str | fileformats.FileFormat | None = None) -> Property: ...
```
ketiltrout:
```suggestion
def file_format(default: str | None = None) -> Property: ...
```
See above.

caput/pipeline/extensions/_extensions.py
```diff
@@ -421,15 +429,37 @@ class BasicContMixin:
     _comm = None
 
     def read_input(self, filename):
-        """Method for reading hdf5 input."""
+        """Method for reading hdf5 input.
```
ketiltrout:
```suggestion
        """Method for reading HDF5 input.
```

caput/pipeline/extensions/_pytools.py
```diff
@@ -15,7 +23,13 @@ def __init__(self, inputs=None):
         self._iter = None
 
     def next(self):
-        """Pop and return the first element of inputs."""
+        """Pop and return the first element of inputs.
+
+        Raises
+        ------
+        :py:exc:`~..exceptions.PipelineStopIteration`
+            When there are no more inputs to provide.
```
ketiltrout:
```suggestion
            No more inputs to provide.
```

caput/pipeline/runner/_scheduler/__init__.py
```diff
@@ -87,24 +91,23 @@ def template_queue(templatefile, var, overwrite, email=None, mailtype=None):
...
+    lint: bool = True,
+    profile: bool = False,
+    profiler: str | None = "cProfiler",
+    psutil: bool = False,
+    overwrite: str = "never",
+    email: str | None = None,
+    mailtype: str | None = None,
 ):
-    r"""Queue a pipeline on a cluster from the given CONFIGFILE.
+    """Queue a pipeline on a cluster from the given CONFIGFILE.
```
ketiltrout:
```suggestion
    """Queue a pipeline on a cluster from the given `configfile`.
```
Also needs a parameters section.

caput/pipeline/runner/_scheduler/__init__.py
```diff
@@ -115,7 +118,6 @@ def queue(
 
     There are many *optional* keys that control more functionality:
 
-    \b
     ``system``
         A name of the cluster that we are running on, if this is supported
         (currently ``gpc``, ``cedar``, ``fir``), this uses more relevant default
```
ketiltrout:
```suggestion
        The name of the cluster that we are running on. If this is a known system
        (currently ``gpc``, ``cedar``, ``fir``), more relevant defaults are used.
```

caput/pipeline/runner/_core.py
```diff
@@ -1,19 +1,32 @@
...
 
 
-def lint_config(configfile):
+def lint_config(configfile: os.PathLike | tuple[os.PathLike, ...]) -> None:
     """Lint a pipeline config file without running it.
 
     Parameters
     ----------
-    configfile : str or tuple of str
+    configfile : os.PathLike
```
ketiltrout:
```suggestion
    configfile : os.PathLike | tuple[os.PathLike, ...]
```
For consistency.

(Also the render is replacing the `...`, with its type (`Ellipsis`), which doesn't seem correct, but probably not worth figuring out.)

caput/pipeline/runner/_core.py
```diff
@@ -28,8 +41,30 @@ def lint_config(configfile):
         Manager.from_yaml_file(cfg, lint=True)
 
 
-def run_pipeline(configfile, profile, profiler, mpi_abort, psutil):
-    """Run a pipeline immediately from the given `configfile`."""
+def run_pipeline(
+    configfile: str,
```
ketiltrout:
```suggestion
    configfile: os.PathLike,
```
Per docstring

caput/pipeline/runner/_core.py
```diff
@@ -44,20 +79,31 @@ def run_pipeline(configfile, profile, profiler, mpi_abort, psutil):
...
+        Positional arguments to pass to `run_pipeline`.
+    \**kwargs : Any
+        Keyword arguments to pass to `run_pipeline`.
     """
     for tfh_name in _from_template(templatefile, var):
         run_pipeline(tfh_name, *args, **kwargs)
 
 
-def _from_template(templatefile, var):
+def _from_template(templatefile: str, var: Any) -> Generator[str]:
```
ketiltrout: I don't think `templatefile` can be a `str`: A `.read()` call is done on it down on line 128.

(If this type needs to be changed, the change will then need to percolate up to all callers of `_from_template`.)

ketiltrout (commented):
I think I now understand at least _one_ of the issues I've been having with the GitHub UI: if my browser loses the session, then my comments disappear because the "new" diff viewer only shows the first 40 comments (for some reason).  GitHub remembers them and they'll eventually pop into existence once I submit the review, but there's no way for me to see them anymore.

So, probably going to keep these small.  This is for the rest of `pipeline.runner`.

caput/pipeline/tasklib/__init__.py
```diff
@@ -1,10 +1,7 @@
-"""caput.pipeline.tasklib.
-
-Collection of basic tasks.
-"""
+"""Collection of basic pipeline tasks."""
```
ketiltrout:
```suggestion
"""A collection of basic pipeline tasks."""
```

caput/pipeline/tasklib/base.py
```diff
@@ -1,4 +1,11 @@
-"""Core implementation of the caput pipeline task class."""
+"""Base pipeline tasks.
+
+Base pipeline tasks providing standard functionality, like logging,
```
ketiltrout:
```suggestion
Base pipeline tasks provide standard functionality, like logging,
```

caput/pipeline/tasklib/base.py
```diff
@@ -20,16 +36,19 @@ class MPILogFilter(logging.Filter):
...
     ----------
-    add_mpi_info : boolean, optional
+    add_mpi_info : bool, optional
         Add MPI rank/size info to log records that don't already have it.
-    level_rank0 : int
-        Log level for messages from rank=0.
-    level_all : int
-        Log level for messages from all other ranks.
+    level_rank0 : int, optional
+        Log level for messages from rank=0. Default is "INFO".
```
ketiltrout:
```suggestion
        Log level for messages from rank=0. Default is `logging.INFO`.
```
So people don't think it's a string.

caput/pipeline/tasklib/base.py
```diff
@@ -199,31 +237,31 @@ def __init__(self):
...
+    then the input will be read (using :py:meth:`~.ContainerTask.read_input`) from the file
+    `self.input_root + self.input_filename`.  If the input is specified both as
     a filename and as a product key in the pipeline configuration, an error
     will be raised upon initialization.
 
-
-    If the value of :attr:`output_root` is anything other than the string
-    "None" then the output will be written (using :meth:`write_output`) to the
+    If the value of :py:attr:`~.ContainerTask.output_root` is anything other than the string
+    ``None`` then the output will be written (using :meth:`write_output`) to the
```
ketiltrout: Given `output_root` is deprecated, it's probably best to de-emphasise this by moving this sentence down in the parameter description for `output_root`.

caput/pipeline/runner/_core.py
ljgray: Oh, that's annoying. I should figure that out...

caput/pipeline/tasklib/base.py
```diff
@@ -399,7 +436,7 @@ def next(self, *input):
         return output if len(output) > 1 else output[0]
 
     def finish(self):
-        """Should not need to override. Implement `process_finish` instead."""
+        """Should not need to override. Implement :py:meth:`~.ContainerTask.process_finish` instead."""
```
ketiltrout: The class docstring says this can be optionally overriden.

caput/pipeline/runner/_core.py
ljgray: Ah - I think it comes from the `cli` click type. I'm going to change that, but not in this PR.

caput/pipeline/tasklib/base.py
ljgray: I think the class docstring is wrong, so I've changed that

caput/pipeline/tasklib/io.py
```diff
@@ -1,8 +1,10 @@
-"""Basic pipeline I/O tasks.
+r"""Collection of tasks for loading and saving files.
```
ketiltrout:
```suggestion
r"""A collection of tasks for loading and saving files.
```

caput/pipeline/tasklib/io.py
```diff
@@ -77,21 +83,22 @@ def list_of_filelists(files: list[str] | list[list[str]]) -> list[list[str]]:
...
     -------
-    The input file list. Any glob patterns will be flattened to file path string lists.
+    file_list : list[PathLike]
+        The input file list. Any glob patterns will be flattened to file path string lists.
 
     Raises
     ------
-    config.CaputConfigError
+    :py:exc:`~caput.config.CaputConfigError`
         If files has the wrong type or if it refers to a file that doesn't exist.
```
ketiltrout:
```suggestion
        If `files` has the wrong type or if it refers to a file that doesn't exist.
```

caput/pipeline/tasklib/io.py
```diff
@@ -126,23 +133,24 @@ def list_or_glob(files: str | list[str]) -> list[str]:
...
-    path strings.
+    file_list : list[PathLike]
+        The input groups. Any glob patterns in the 'files' list will be flattened to file
+        path strings.
 
     Raises
     ------
-    config.CaputConfigError
+    :py:exc:`~caput.config.CaputConfigError`
         If groups has the wrong format.
```
ketiltrout:
```suggestion
        If `groups` has the wrong format.
```

caput/util/arraytools.py
```diff
@@ -30,18 +47,103 @@ def _array_size(arr: np.ndarray):
...
+    full_list : list
+        The full list to partition.
+    i : int
+        The index of the partition to return.
+    n : int
+        The number of partitions to create.
+    method: {"con", "alt", "rand"}, optional
+        The method to use for partitioning. Options are:
+
+        - "con": contiguous partitionsI
```
ketiltrout:
```suggestion
        - "con": contiguous partitions
```

caput/util/arraytools.py
```diff
@@ -30,18 +47,103 @@ def _array_size(arr: np.ndarray):
...
+        - "alt": alternating partitions
+        - "rand": random partitions
+
+        Default is "con".
+
+    Returns
+    -------
+    partition : list
+        The `i` th partition of the list.
```
ketiltrout:
```suggestion
        The `i`th partition of the list.
```

caput/util/arraytools.py
```diff
@@ -30,18 +47,103 @@ def _array_size(arr: np.ndarray):
...
+
+
 def unique_ordered(x: Iterable) -> list:
     """Take unique values from an iterable with order preserved.
 
     Parameters
     ----------
     x : Iterable
-        An iterable to get unique values from
+        An iterable which may or may not container duplicate values.
```
ketiltrout:
```suggestion
        An iterable which may or may not contain duplicate values.
```

caput/util/mpitools.py
```diff
@@ -1,9 +1,9 @@
 """Utilities for making MPI usage transparent.
 
 This module exposes much of the functionality of :mod:`mpi4py` but will still
-run in serial if mpi is not present on the system. It is thus useful for
+run in serial if MPI is not present on the system. It is thus useful for
```
ketiltrout:
```suggestion
run in serial if MPI is not present on the system. It is, thus, useful for
```

caput/util/mpitools.py
```diff
@@ -1,9 +1,9 @@
...
 writing code that can be run in either parallel or serial. Also it exposes all
-attributes of the :mod:`mpi4py.MPI` by the :class:`SelfWrapper` class for
+attributes of the :mod:`mpi4py.MPI` by the :class:`_SelfWrapper` class for
```
ketiltrout:
```suggestion
attributes of the :mod:`mpi4py.MPI` module by the :class:`_SelfWrapper` class for
```

caput/util/mpitools.py
```diff
@@ -293,293 +309,291 @@ def available_memory_shared() -> int:
...
 
-    Returns
-    -------
-    results : list
-        Global list of results.
-    """
-    # Synchronize
-    barrier(comm=comm)
+def active_comm(aprocs: list) -> MPI.Comm | None:
+    """Return a communicator consists of a list of processes in `aprocs`."""
```
ketiltrout:
```suggestion
    """Return a communicator consisting of processes listed in `aprocs`."""
```

(Also, probably all of these functions that take and `aprocs` should have a Parameters section.)

caput/util/mpitools.py
```diff
@@ -293,293 +309,291 @@ def available_memory_shared() -> int:
...
+    # create a new communicator from active processes
+    return _comm.Create(_comm.Get_group().Incl(aprocs))
 
-    # Pair up each list item with its position.
-    zlist = list(enumerate(glist))
 
-    # Partition list based on MPI rank
-    llist = partition_list_mpi(zlist, method=method, comm=comm)
+def active(aprocs: list) -> MPI.Comm | None:
+    """Make a list of processes in `aprocs` active, while others wait."""
```
ketiltrout:
```suggestion
    """Make processes listed in `aprocs` active, while others wait."""
```

caput/util/mpitools.py
```diff
@@ -293,293 +309,291 @@ def available_memory_shared() -> int:
...
     Parameters
     ----------
-    dtype : np.dtype
-        The numpy datatype.
+    full_list : list
+        The full list to partition.
+    method : {"con", "alt", "rand"}, optional
+        How to split the list, can be "con": contiguous, "alt": alternating,
+        "rand": random. Default is "con".
+    comm : MPI.Comm | None
```
ketiltrout:
```suggestion
    comm : MPI.Comm | None, optional
```

caput/util/mpitools.py
```diff
@@ -293,293 +309,291 @@ def available_memory_shared() -> int:
...
+    method = kargs.get("method", "con")
+    comm = kargs.get("comm", _comm)
 
-    return np.array([part, bound[:m], bound[1 : (m + 1)]])
+    return partition_list_mpi(full_list, method=method, comm=comm)
 
 
-def split_all(n, comm=_comm):
+def split_all(n: int, comm: MPI.Comm | None = _comm) -> np.ndarray:
     """Split a range (0, n-1) into sub-ranges for each MPI Process.
```
ketiltrout: See my comment in `arraytools` about proper brackets for the range.

caput/util/mpitools.py
```diff
@@ -738,32 +827,42 @@ def transpose_blocks(row_array, shape, comm=_comm):
...
+    fname : str
         Name of the file to write.
-    dsetname : string
+    dsetname : str
         Name of the dataset to write (must be at root level).
-    shape : tuple
+    shape : tuple[int, ...]
         Shape of the dataset.
-    dtype : numpy datatype
+    dtype : dtype
```
ketiltrout: Not `npt.DTypeLike`?

ljgray: Within docstrings, intersphinx mapping seems to find `dtype`, `array_like`, and a few other types like that. Although, ~5% of the time this linking seems to fail and I don't know why. `npt.DtypeLike` is only used for type hints. It's annoyingly confusing.

caput/util/profiler.py
```diff
@@ -172,35 +180,35 @@ class PSUtilProfiler(psutil.Process):
...
     Parameters
     ----------
     use_profiler : bool
         Whether to run the profiler or not.
-    label : str
+    label : str, optional
         Default description of what is being profiled.
-    logger
+    logger : logging.Logger | None, optional
         If a logging object is passed the values of the IO done counters are logged
```
ketiltrout:
```suggestion
        If not `None`, the values of the IO done counters will be logged
```

caput/util/profiler.py
```diff
@@ -318,30 +325,12 @@ def start(self):
         self._monitor_thread = threading.Thread(target=self.monitor)
         self._monitor_thread.start()
 
-    def stop(self):
-        """Stop profiler. Dump results to csv file and/or log and/or set results on self.usage.
+    def stop(self) -> None:
+        """Stop profiler.
```
ketiltrout:
```suggestion
        """Stop profiling.
```
For consistency with `start`.

caput/util/profiler.py
```diff
@@ -446,21 +435,43 @@ def bytes2human(num):
...
+
+        Returns
+        -------
+        usage : dict
+            Usage dictionary with the following keys:
+
+            ``cpu_times`` : dict
+                dict version of `psutil.cpu_times`. Process CPU times since `start` was called in seconds.
+            ``cpu_percent`` :  float
+                Process CPU utilization since `start` was called as percentage. Can be >100 if multiple threads run on
```
ketiltrout:
```suggestion
                Process CPU utilization since `start` was called as a percentage. Can be >100 if multiple threads run on
```

caput/util/truncate.pyi
```diff
@@ -0,0 +1,152 @@
...
+    truncated : array_like
+        The modified array. This shares the same underlying memory as the input.
+    """
+    ...
+
+#
+def bit_truncate_max_complex(
+    val: np.ndarray[np.complex128], prec: np.float32, prec_max_row: np.float32
+) -> np.ndarray[np.complex128]:
+    """Truncate using a relative per element and per the maximum of the last dimension.
```
ketiltrout:
```suggestion
    """Truncate using a relative per-element and per the maximum of the last dimension.
```

caput/util/truncate.pyi
```diff
@@ -0,0 +1,152 @@
...
+
+    This scheme allows elements to be truncated based on their own value and a
+    measure of their relative importance compared to other elements. In practice the
+    per element absolute precision for an element `val[i, j]` is given by `max(prec *
```
ketiltrout:
```suggestion
    per-element absolute precision for an element ``val[i, j]`` is given by ``max(prec *
```

caput/util/truncate.pyi
```diff
@@ -0,0 +1,152 @@
...
+    val[i, j], prec_max_dim * val[i].max())`
```
ketiltrout:
```suggestion
    val[i, j], prec_max_dim * val[i].max())``.
```

ketiltrout (commented):
The rest of `util`.

caput/config.py
```diff
@@ -264,55 +290,54 @@ def _finalise_config(self):
         To be overridden in subclasses if we need to perform some processing
         post configuration.
         """
+        ...
 
 
-def utc_time(default=None):
+def utc_time(default: TimeLike | None = None) -> Property[TimeLike]:
     """Property for representing UTC as UNIX time.
```
ketiltrout:
```suggestion
    """Property type for representing UTC as UNIX time.
```
For consistency with other `Property`-returning functions.

caput/config.py
```diff
@@ -368,23 +392,28 @@ def _prop(val):
...
+    maxlength: int | None = None,
+    default: Any | None = None,
+) -> Property:
     """Property type that validates lists against required properties.
 
     Parameters
     ----------
-    type_ : type, optional
+    type\\_ : dtype | None, optional
         Type to apply. If `None` does not attempt to validate elements against type.
```
ketiltrout:
```suggestion
        Required element type. If `None`, no type validation is done.
```

caput/config.py
```diff
@@ -436,28 +464,27 @@ def _prop(val):
...
     """Property type that validates the caput logging config.
 
     Allows the type to be either a string (for backward compatibility) or a dict
     setting log levels per module.
 
     Parameters
     ----------
-    default : optional
+    default : str | dict | None, optional
         The optional default value.
```
ketiltrout:
```suggestion
        The optional default logging config.  A simple string level is applied to the root logger.
```

caput/mpiarray.py
```diff
@@ -1,4 +1,9 @@
-"""An array class for containing MPI distributed array.
+"""MPI-distribued numpy array.
+
+:py:mod:`~caput.mpiarray` provides a distributed array object which is a subclass of
+:py:class:`numpy.ndarray`. The goal of :py:mod:`~caput.mpiarray` is to enable seamless
```
ketiltrout:
```suggestion
This module provides a distributed array object which is a subclass of
:py:class:`numpy.ndarray`. The goal of this module is to enable seamless
```
These are self-links which aren't helpful.

caput/mpiarray.py
```diff
@@ -83,39 +88,39 @@
...
 >>> for ri in range(mpitools.size):
-...    if ri == mpitools.rank:
-...        print(ri, arr2)
-...    mpitools.barrier()
+...     if ri == mpitools.rank:
+...         print(ri, arr2)
+...     mpitools.barrier()
 0 [0.]
 
 Fetch a view of the whole array with a partial slice. The final two ranks should be None
```
ketiltrout:
```suggestion
Fetch a view of the whole array with a partial slice. The final two ranks should be None

```
Blank line needed to properly format the code block.

caput/mpiarray.py
```diff
@@ -83,39 +88,39 @@
...
+...         print(ri, arr3)
+...     mpitools.barrier()
 0 [0.]
 
 Direct Slicing
 ==============
 
-`MPIArray` supports direct slicing using `[...]` (implemented via
-:py:meth:`__getitem__` ). This can be used for both fetching and assignment. It is
+:py:class:`.MPIArray` supports direct slicing using `[...]` (implemented via
```
ketiltrout:
```suggestion
:py:class:`.MPIArray` supports direct slicing using ``[...]`` (implemented via
```

caput/mpiarray.py
```diff
@@ -83,39 +88,39 @@
...
+:py:meth:`.MPIArray.__getitem__` ). This can be used for both fetching and assignment. It is
```
ketiltrout:
```suggestion
:py:meth:`.MPIArray.__getitem__`). This can be used for both fetching and assignment. It is
```

caput/mpiarray.py
```diff
@@ -83,39 +88,39 @@
...
 recommended to only index into the non-parallel axis or to do a full slice `[:]`.
```
ketiltrout:
```suggestion
recommended to only index into the non-parallel axis or to do a full slice ``[:]``.
```

caput/mpiarray.py
```diff
@@ -83,39 +88,39 @@
...
 
 Direct Slicing Behaviour
 ------------------------
 
-- A full slice `[:]` will return a :class:`MPIArray` on fetching, with
+- A full slice `[:]` will return a :py:class:`~.MPIArray` on fetching, with
```
ketiltrout:
```suggestion
- A full slice ``[:]`` will return a :py:class:`~.MPIArray` on fetching, with
```

caput/mpiarray.py
```diff
@@ -667,39 +678,41 @@ def _view_from_data_and_params(
...
-        an `np.ndarray`.
+    def wrap(
+        cls, array: npt.NDArray, axis: int, comm: MPI.Comm | None = None
+    ) -> MPIArray:
+        """Turn a group of :py:class:`~np.ndarray` into a distributed :py:class:`.MPIArray` object.
 
         Parameters
         ----------
-        array : np.ndarray
+        array : MPIArray
```
ketiltrout: The type hint is `npt.NDArray`

caput/mpiarray.py
```diff
@@ -843,38 +864,37 @@ def allreduce(self, op=None):
...
         ----------
-        axis : integer
-            Which access to enumerate over.
+        axis : int
+            Axis to enumerate over.
 
         Returns
         -------
-        iterator : (local_index, global_index)
+        enumerator : Iterator
```
ketiltrout:
```suggestion
        enumerator : enumerate
```
`enumerate()` is a class.  This function is returning an instance of it.

caput/mpiarray.py
```diff
@@ -884,62 +904,88 @@ def enumerate(self, axis):
...
             to select the most efficient axis for the reading.
-        sel : tuple, optional
+        sel : SelectionTupleLike | None, optional
             A tuple of slice objects used to make a selection from the array
             *before* reading. The output will be this selection from the dataset
             distributed over the given axis.
-        file_format : `fileformats.HDF5` or `fileformats.Zarr`
-            File format to use. Default `fileformats.HDF5`.
+        file_format : fileformat.FileFormat, optional
+            File format to use. Default :py:class:`~caput.fileformats.HDF5`.
```
ketiltrout: Not linked

caput/mpiarray.py
```diff
@@ -1042,30 +1088,36 @@ def from_file(
...
             Name or identifier of HDF5 compression filter.
-        compression_opts
+        compression_opts : tuple | None, optional
             See HDF5 documentation for compression filters.
             Compression options for the dataset.
+
+        Raises
+        ------
+        ValueError
+            If :py:mod:`h5py` is built without MPI support and the filename
```
ketiltrout:
```suggestion
            If :py:mod:`h5py` is built without MPI support and `f`
```

caput/mpiarray.py
```diff
@@ -1241,121 +1303,122 @@ def to_zarr(
...
         Parameters
         ----------
-        axes : None, tuple of ints, or n ints
+        \*axes : Sequence[int] | int | None, optional
+            Transpose arguments:
+
             - None or no argument: reverses the order of the axes.
-            - tuple of ints: i in the j-th place in the tuple means a`s i-th axis
+            - sequence of ints: i in the j-th place in the tuple means a`s i-th axis
               becomes a.transpose()`s j-th axis.
```
ketiltrout:
```suggestion
            - sequence of ints: `i` in the `j`th place in the tuple means `a`'s `i`th axis
              becomes ``a.transpose()``'s `j`th axis.
```
Not really sure how best to format this.  Maybe more double backticks would be better?

caput/mpiarray.py
```diff
@@ -1509,13 +1581,14 @@ def gather(self, rank=0):
...
-    def allgather(self):
+    def allgather(self) -> np.ndarray:
         """Gather a full copy onto each rank.
 
         Returns
         -------
-        arr : np.ndarray
-            The full global array.
+        gathered_array : ndarray
+            The full global array gather into a :py:class:`~np.ndarray`
```
ketiltrout:
```suggestion
            The full global array gathered into a :py:class:`~np.ndarray`
```

caput/mpiarray.py
```diff
@@ -1714,40 +1791,58 @@ def __array_ufunc__(
...
+          otherwise there are no restrictions (e.g. `np.cumsum`).
         - "reduce": the reduction cannot take place along the distributed axis (i.e.
-        `axis` must not include the distributed axis), *except* for a total reduction
-        over all axes (`axis = None`) which will return an array with a single element
-        on each rank.
+          `axis` must not include the distributed axis), *except* for a total reduction
+          over all axes (`axis = None`) which will return an array with a single element
+          on each rank.
 
         For all of these method types `out` keyword arguments, used for directly placing
```
ketiltrout:
```suggestion
        For all of these method types, `out` keyword arguments, used for directly placing
```

caput/mpiarray.py
```diff
@@ -1714,40 +1791,58 @@ def __array_ufunc__(
...
-        finally using an `MPIArray.wrap` call to construct a distributed array from the
+        :py:attr:`~.MPIArray.local_array` to get the local numpy data, then applying the ufunc, and
+        finally using an :py:meth:`~.MPIArray.wrap` call to construct a distributed array from the
         output.
 
         Parameters
         ----------
-        ufunc: <function>
+        ufunc : ufunc
             ufunc object that was called
```
ketiltrout:
```suggestion
            ufunc object that to call
```

doc/conf.py
```diff
@@ -317,5 +616,4 @@ def setup(app):
 # If false, no module index is generated.
 # texinfo_domain_indices = True
 
-# How to display URL addresses: 'footnote', 'no', or 'inline'.
-# texinfo_show_urls = 'footnote'
+# How to display URL addresses: 'footnote', 'no', or 'inline'. texinfo_show_urls = 'footnote'
```
ketiltrout:
```suggestion
# How to display URL addresses: 'footnote', 'no', or 'inline'.
# texinfo_show_urls = 'footnote'
```
Restore a dropped newline. (To avoid potential confusion in the future.)

caput/config.py
```diff
@@ -190,36 +207,45 @@ class Reader:
     """A class that allows the values of Properties to be assigned from a dictionary."""
 
     @classmethod
-    def from_config(cls, config, *args, **kwargs):
-        """Create a new instance with values loaded from config.
+    def from_config(cls, config: dict, *args: Any, **kwargs: Any) -> Reader:  # noqa: D417
+        r"""Create a new instance with values loaded from config.
```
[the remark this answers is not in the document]
ljgray: I think this is because all of the `Property` methods are private. I think that `from_config` should probably be public, so I'll have to go and find all cases of this use in a future PR

PR #313, opened 2025-12-10
Update base container API

ljgray:
This should hopefully be the final API change.

- Changes `memdata.BasicCont` to `containers.Container`. Functionality is unchanged.
- Changes `containers.ContainerBase` to `containers.ContainerSpec` to try to better convey that the only additional functionality over `Container` is the dataset and axes specs. Same change for `TableBase`.


ketiltrout:
If the erstwhile `BasicCont` _is_ the nominal minimal container, I'd okay with it being `Container`.

What about `DatasetContainer` for the second?   Or something like `ContainerProto` to emphasize it's a prototype class?

PR #315, opened 2025-12-17
Use more explicit thread mapping with mpirun

ljgray:
The way that `srun` maps threads to cores seems to occasionally result in some processes running significantly slower than others. I'm not entirely sure why, but explicitly binding threads to the L3 cache to take advantage of Fir's numa layout seems to fix this.


ketiltrout (approved):
Hmm.  I've just discovered that "close" isn't a good value for a variable because it's spelled the same as "close".

PR #316, opened 2026-01-07
MAJOR: package refactor

ljgray:
This is a major refactor of the package API.

_Note to reviewers: The easiest way to review this is probably just to switch to this branch and explore, either through github or locally. The diff will be a nightmare._

# API
## Algorithms
Submodule for fast implementations of various algorithms. This mostly contains things already in `caput`, but ports `random.MultithreadedRNG` from `draco`.

## Astro
Submodule for astro-related things. Includes time conversions, coordinate transforms, the `Observer` class, some constants, and a skyfield wrapper. This mostly already existed in `caput`, but a few coordinate transforms were ported from `cora`.

## Config
API is unchanged.

## Containers
Ports the base `Container` infrastructure from `draco`, along with a few generic base-level containers. The API is exposed via `__init__.py`.

## Memdata
Exposes the `memh5` and `fileformat` interfaces. This is intended to provide a standard API to facilitate a future rework of `memh5`.

## MPIarray
API is unchanged.

## Pipeline
The `caput-pipeline` infrastructure. The basic blocks are provided as `Manager` and `Task`, while a collection of basic tasks are provided in `tasklib`. Importantly, the widely-used `SingleTask` is renamed to `tasklib.base.ContainerTask`.

## Util
Some helpful utilities.

# Docs and Typing
This PR also adds extensive type hinting support and fully-developed API reference documentation, which is linked in the README. In more complex cases, typing as provided via `.pyi` stub files to avoid polluting the actual code (`.py`). There doesn't _really_ seem to be a standard for using stub files vs. adding typing directly, so I did this on a case-by-case basis. In theory, we could change this in the future without affecting the user.

# Other
The main goal of this PR is to provide an updated, stable API to ease future backend reworks and to reduce the dependencies required for other packages to run pipelines and use distributed containers. I don't want to make any **Major** API changes at this point, but I'm happy to hear feedback and to make some final tweaks.


sjforeman:
@ljgray and @ketiltrout , congratulations on the heroic work that went into this! I skimmed through a bunch of it and don't have any objections.

If you're interested in moving other things in this PR, I'd be in favor of moving `cora.util.cubicspline` into `caput.util`, and `cora.signal.lsscontainers.InterpolatedFunction` into `caput.containers.util` (not sure where else it should go). However, if you'd just like to get everything merged for rev09, those can wait until later.

PR #318, opened 2026-01-13
Update all unit tests to work with MPI

ljgray:


README.md
```diff
@@ -52,3 +52,9 @@ At minimum, all new code must include type hints for:
 - Global variables
 
 Additional type hints are encouraged but not required.
+
+### Tests
+All new code should have corresponding unit tests which pass both with and without MPI.
+Tests which are MPI-specific (or are expected to fail with MPI) should be marked using
+`pytest.mark.mpi`, `pytest.mark.mpi_xfail`, or `pytest.mark.mpi_skip`. Tests run with
+MPI should include the flag `--with-mpi` in order for marks to be set correctly.
```
ketiltrout:
```suggestion
Tests which should only be run with MPI should be marked using `pytest.mark.mpi`.
Tests that are expected to xfail when run with MPI should be marked with
`pytest.mark.mpi_xfail`.  Finally, tests that should be skipped when running
with MPI should use `pytest.mark.mpi_skip`. When running test with MPI, the
invocation should include the flag `--with-mpi` in order for these marks to be
set correctly.
```

and/or just link to https://pytest-mpi.readthedocs.io/en/latest/markers.html

PR #320, opened 2026-01-30
Include pre-commit `ruff` hooks

ljgray:


ketiltrout:
> Is there an appropriate place to add `ruff` as a dependency? It's not a runtime dependency. It's not really a build dependency, so ....?

Oh, wait, it is there already in the dev dependencies.  Nevermind.

PR #340, opened 2026-04-22
feat(runner): allow profiling of template jobs

sjforeman:
This PR is motivated by a bug that @nKrugo13 identified in the refactored pipeline routines: `caput-pipeline template-run` will throw an error if `--submit` is not specified and any of the `profile`, `profiler`, `mpi_abort`, or `psutil` arguments are missing. This happens because `_cli.template_run()` calls `_core.template_run()`, which in turn calls `_core.run_pipeline()`, for which the 4 arguments above are not optional. I believe this bug was introduced in the process of decoupling the command-line interface from `run_pipeline()`: previously, `run_pipeline()` was called directly from the command line, and always had default values for these 4 arguments provided.

In the process of fixing this bug, I realized that it would be straightforward to allow profiling of `template-run` jobs, so I've implemented that here.

### DIFF
--- a/caput/pipeline/runner/_cli.py
+++ b/caput/pipeline/runner/_cli.py
@@ -21,6 +21,60 @@ def cli():
     pass
 
 
+def profiler_options(f):
+    """Options for profiling a pipeline run."""
+    options = [
+        click.option(
+            "--profile",
+            is_flag=True,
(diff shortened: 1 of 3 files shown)


ljgray (approved):
This is great

PR #345, opened 2026-07-13
Pipeline: add an optional 'if' and 'after' parameters for conditional tasks and better control flow

ljgray:
# if
Useful for templates pipelines where we might want to skip a task under certain conditions, without messing up key inheritance. The pipeline will internally replace the conditional task with a `NoOp` if the conditional is `False`, in which case data keys are simply passed through. Naturally, this will fail for a task which produces a different number of outputs than it takes as input, or which does not intend to pass through the input, so this can likely be made to be more robust.

# after
Effectively just streamlines the `WaitUntil` task - adds a set of data product keys (similar to `in`, `out`, and `requires`) which indicate data products that must exist before a task can run, but are not actually consumed by the task.

# IO
Adds two extra tasks:
- Simple task to load _all_ files from a list at once
- More complicated task to load batches of files together. This ends up being necessary because the pipeline never seems producer-only tasks as being "available", so they are simply run in order when there is nothing else to do. However, the order _always_ starts at the first possible task, so two back-to-back IO tasks will result in the pipeline first loading every file for the first task before proceeding.


ketiltrout (approved):
This all seems reasonable.

> Naturally, this will fail for a task which produces a different number of outputs than it takes as input, or which does not intend to pass through the input, so this can likely be made to be more robust.

I don't really think an `if` makes sense unless the number of inputs equals the number of outputs because the person creating the parameter has no way to explain to the pipeline what to do in the False case.  I'd say encountering an `if` on a task which whose input and output counts differ should just raise a `CaputConfigError` or something.

A more robust solution for conditional pipeline flow would need to involve specifying a conditional along with two tasks, one to run when the conditional was true and the other to run if it were False.  Or maybe that's just better done within a task, rather than at the pipeline level.

ljgray:
> I don't really think an `if` makes sense unless the number of inputs equals the number of outputs because the person creating the parameter has no way to explain to the pipeline what to do in the False case. I'd say encountering an `if` on a task which whose input and output counts differ should just raise a `CaputConfigError` or something.

I generally agree, but one goal is to be able to skip entire sequences of tasks, rather than just a single individual one - it's possible that some sequence on tasks could produce intermediate products. I think the workaround here is to implement #347, and just require that a conditional _block_ has the same number of inputs/outputs.

